PR #2, opened 2021-09-20
Organize old wiki docs into folders

sandy-wright:
# Description

I created a simple folder structure in the `docs/` directory, and moved the docs from the wiki to the folder I thought was most appropriate. I also added a couple of index pages that have not yet been fleshed out. I didn't change anything in the docs other than that! So it looks huge, but it was just shuffling a bunch of things.

Please also note that I added a `Vault` directory. This will serve as an archive folder. We can move docs here when we're not sure where they belong, or we aren't sure if they're still accurate or needed. Right now, it contains documents that I'm sure should be moved to Confluence (in the `docs/vault/confluence` dir) and infra documentation that I'm not sure should be moved, but I _think_ probably should also be in Confluence (in the `docs/vault/infra` dir).

This won't be deployed to github pages, so you'll need to run it locally to see the changes. I also **highly recommend doing this** because it'll be almost impossible to actually absorb the changes I made unless you experience it in the app.

To do this, run:

```sh
yarn install
yarn start
```


rogeruiz:
This looks great @sandy-wright! I'm wondering if for docs that go in `Vault` if it [makes sense to **version** the documentation](https://docusaurus.io/docs/versioning) or rename the folder to `Deprecated` instead?

I don't think answering the question should stop this from getting merged in. But I'm curious what your thoughts are on this.

scarequotes:
Thank you for all the Terminal setup @sandy-wright — looks great to me!

PR #6, opened 2021-09-23
Add in documentation for how to use Redocusaurus

rogeruiz:
This documentation exists outside of the Docusaurus site because it's
about configuring this site. I think eventually though it may exist
inside of the Docusaurus site under a directory named `Meta` or
something similar as it's about how to operate and maintain this
website.

This came up in our last Prime API Documentation meeting, and I'm 
completing it as part of the action items in that meeting. I'm tagging 
folks that were in this meeting for review.


sandy-wright (approved):
Thanks for adding this!

PR #13, opened 2021-10-06
Add Nix setup and improve README instructions

monfresh:
Since Nix has proven useful on other projects, I thought I'd add it as an
option here. I plan on submitting a separate PR that automates much of
the manual setup for both the Nix and Brew options via a shell script.﻿


ahobson (approved):
Thank you so much for this!

PR #22, opened 2021-10-19
[MB-9756] Update location of new top-level apperror package

hannaliebl:
Changes introduced in https://github.com/transcom/mymove/pull/7593 moved the errors that used to be in pkg/services/errors to a top-level apperror package.


reggieriser (approved):
LGTM!  Thanks for fixing this!

PR #24, opened 2021-10-20
[MB-9609] Migrate Backend all pages

rogeruiz:
This patch migrates all the Backend ~~Testing~~ pages to its own section. Because of how the docs are setup in Docusaurus, they will start to appear on the top left so they are not integrated into “Docs” while we migrate more pages.

This PR is an amalgamation of other PRs, #27 #30.

PR #25, opened 2021-10-20
Add page redirects

rogeruiz:
### Description

This patch series adds the [Client Redirects plugin to
Docusaurus](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects/).
With so much of the documentation changing location now, a redirect plugin will
be useful so folks using the documentation don't encounter 404 errors. I don't
think this is a long-term solution, but these redirects are useful when
documentation is changing location.

- [x] Rename some about pages to lowercase
- [x] Add in plugin-client-redirects
- [x] Add in examples for lowercase renaming redirects


rogeruiz:
This should be merged in before #24 so that we can redirect the Backend Testing pages over to the new structure.

PR #27, opened 2021-10-21
[MB-9609] Migrate Backend Guides pages

rogeruiz:
This patch migrates all the Backend Guides pages to its own section. Because of how the docs are setup in Docusaurus, they will start to appear on the top left so they are not integrated into “Docs” while we migrate more pages.

The PR builds on the work of #24 and must be merged before that PR can be merged into `main`.


docs/api/guides/guide-to-creating-an-endpoint.md
```diff
@@ -275,3 +279,4 @@ var eventModels = map[KeyType]eventModel{
 
 
 If you'd like to learn more about event triggers, you can find more details [here](https://github.com/transcom/mymove-docs/blob/720592c63db4bffe402a801417f7c14772573c28/docs/dev/contributing/backend/How-to-Add-an-Event-Trigger.md).
+
```
gidjin, line 282: Sometimes I wonder if we should standardize on adding or not adding a newline at the end of the file.

rogeruiz:
Merging this in myself since it's not being merged into `main` directly.

PR #47, opened 2021-11-10
Clarify GOROOT troubleshooting for asdf and don't hardcode env value

duncan-truss:
While trying to run the standalone ato-linter tool I was having what looked like go build/version issues with types and packages.  I knew we were on go version 1.17.2 so it was weird to see references to an older go version that wasn't pointing to the asdf install.

`pkg/services/mocks/SFTPClient.go:6:2: package io/fs is not in GOROOT (/usr/local/Cellar/go@1.14/1.14.7/libexec/src/io/fs)`

Upon further research I saw that my GOROOT was not being set to the asdf go location so I reshimed and updated my GOROOT with `asdf reshim golang && export GOROOT="$(asdf where golang)/go/"`

I am going to add GOROOT to my ~/.zshrc file going forward with the output from asdf as per instructions.

### DIFF
--- a/docs/backend/guides/how-to/manage-golang-with-asdf.md
+++ b/docs/backend/guides/how-to/manage-golang-with-asdf.md
@@ -91,7 +91,7 @@ OK: all prereqs found
 
 ### GOROOT is showing a different version or not working with GoLand
 
-Another issue GoLand users and others can run into with asdf is with GOROOT. Goland expects GOROOT to be set but it may be unclear what it should be set to. You can easily find the value by running `asdf where go`. Not upgrading versions will require this to change unless you can set GOROOT to the output directory of `asdf where go`. For example, you can set this in your `~/.zshrc` (or similar) shell:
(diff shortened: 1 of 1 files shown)


felipe-lee (commented):
Just to make sure I understand fully, the `asdf reshim golang` part was only needed for you because you did it directly as a command, but having `export GOROOT="$(asdf where golang)/go/"` in your .*rc file will be enough?

Also, maybe we should update the wording here:
> running `asdf where go` and then set that appropriately for your shell

To explicitly say to set it in the .*rc files?

duncan-truss:
> Just to make sure I understand fully, the `asdf reshim golang` part was only needed for you because you did it directly as a command, but having `export GOROOT="$(asdf where golang)/go/"` in your .*rc file will be enough?
> 
> Also, maybe we should update the wording here:
> 
> > running `asdf where go` and then set that appropriately for your shell
> 
> To explicitly say to set it in the .*rc files?

Yea I don't believe the `asdf reshim` was necessary since I didn't install golang outside of the asdf plugin https://asdf-vm.com/manage/core.html#reshim .  That would be solving a different problem.

felipe-lee (approved):
Awesome, LGTM, thanks for updating our docs!

PR #48, opened 2021-11-18
Update docs to reflect new http_route

ronaktruss:
Updated docs for local telemetry to reflect the changes made in this [PR](https://github.com/transcom/mymove/pull/7745).


docs/tools/telemetry/running-telemetry-locally.md
```diff
@@ -29,15 +29,21 @@ Running the elastic search offering locally turns out to be easy.
...
-1. To view this data, on the Kibana page you can click the `Observability` panel and then navigate to the `APM` -> `Services` section on the left hand navigation bar.
+1. To view this data, on the Kibana page you can click the `Observability` panel and then navigate to the `APM` -> `Services` section on the left hand navigation bar. Then click on the `milmove` service.
 
   ![Services Dashboard](/img/telemetry/services-dashboard.png)
 
-  :::info
+## How to find specific transactions
 
-  Services are all named `unknown_service_milmove_gin` and all local transactions besides logging in and logging out are called `server-no-tls`. This is a known [bug](https://dp3.atlassian.net/browse/MB-9926).
+To find specific transactions you can leverage the search bar and utilize some of the metadata added to each transaction.
```
duncan-truss, line 38: This goes back to what is the right verbiage, I think technically we are filtering traces within the server-no-tls transaction(s)?  I don't know if AWS has a first class concept of transactions the way they are in Kibana.  No action needed.

I saw in the otel docs they recommend setting the trace/span name to the http route https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span.  We should probably try conforming to that next.

docs/tools/telemetry/running-telemetry-locally.md
```diff
@@ -29,15 +29,21 @@ Running the elastic search offering locally turns out to be easy.
...
 
-  :::
+Here are a list of some of the metadata fields that you can search against and what information is provided in that field:
+
+| Key                       | Description                             | Example                                     |
+|---------------------------|-----------------------------------------|---------------------------------------------|
+| labels.http_route         | Shows the non dynamic url that is hit   | /internal/service_members/{serviceMemberId} |
+| http.response.status_code | Show the corresponding http status code | 200                                         |
+| http.request.method       | Shows the http request method           | PATCH                                       |
```
duncan-truss, line 46: Yea this is a good point since the route alone wouldn't distinguish between different methods on the same route.

duncan-truss (approved):
Thanks for putting a bow on this

PR #50, opened 2021-11-29
Add reference to how to test notifications

jacquelineIO:
Add reference to how to test notifications

### DIFF
--- a/docs/backend/guides/how-to/add-an-event-trigger.md
+++ b/docs/backend/guides/how-to/add-an-event-trigger.md
@@ -102,4 +102,4 @@ var ghcEndpoints = EndpointMapType{
 After that, you have a new endpoint you can pass in with your event trigger.
 
 ## Reference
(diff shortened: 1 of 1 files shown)


docs/backend/guides/how-to/add-an-event-trigger.md
```diff
@@ -100,3 +100,6 @@ var ghcEndpoints = EndpointMapType{
 ```
 
 After that, you have a new endpoint you can pass in with your event trigger.
+
+## Reference
+* [Acceptance testing notifications](https://transcom.github.io/mymove-docs/docs/backend/testing/acceptance-testing-notifications/)
```
rogeruiz, line 105:
```suggestion
* [Acceptance testing notifications](../../testing/acceptance-testing-notifications.md)
```

rogeruiz (approved):
lgtm 🌈 Thanks @jacquelineIO I'll approve this, but I also think it might be a good idea to do the links like this so that Docusaurus knows that they're internal to the site and not external. 

But this won't do things like catch broken links right now since we only warn on broken links.

PR #61, opened 2022-01-06
Update the Vale configuration

rogeruiz:
This patch series should not get too big. And I want to attempt to only update the Vale configuration and leave the documentation alone. When it comes to updating documentation. I will open individual PRs for just that single document so that folks aren't having to review a large PR.

So this PR is just the beginning of a lot of little PRs.


rogeruiz:
After some more thoughts on this, this PR is going to be more about splitting up the files found the `vale/Vocab` directory so that it's easier to maintain going forward.

rogeruiz:
Created about 43 branches which have been validated as I make my way through the documentation running Vale on each file. I have about 85 files left that still trigger warnings or suggestions given the current configuration for Vale.

rogeruiz:
The error in this PR is because we're not using the `HEAD` version of Vale. We can either way for a release, or supply our own installation of Vale in a GitHub Action to fix this. The author mentioned that it would be released soon in the source repository.

```
Error: {
"Code": "E100",
"Text": "E100 [vocab] Runtime error\n\nVocab 'Software, Truss, Transcom' does not exist\n\nExecution stopped with code 1."
}
```

[=> Error about multiple Vocabs](https://github.com/transcom/mymove-docs/runs/4783970651?check_suite_focus=true#step:4:9)
[=> Author stating that a new release would be coming this week](https://github.com/errata-ai/vale/issues/241#issuecomment-1007878583)

YanZ777 (approved):

vale/Vocab/Software/accept.txt
```diff
@@ -0,0 +1,135 @@
...
+(?i)PostgreSQL
+Syncada
+nullable
+(?i)etag(s)?
+env
+dev
+(?i)classname(s)?
+repo
+autogenerate
+Transcom
```
ronaktruss, line 20: This also exists in the `Transcom` vocab

ronaktruss (commented):
I have 2 general questions:
1. What is the purpose of having multiple vocab files?
    - The main benefit I see is having some context what specific words mean should I be brand new to this project. The downside is a continual effort is needed to ensure that words are added to the correct context/file.
2. Is the purpose of this PR to just have this run/setup on the CI side? Is the expectation for people to use this locally using the cli tool? Do we need to add instructions on how to do so?

PR #85, opened 2022-01-12
[vale] Review docs Managing Golang with asdf

rogeruiz:
This PR fixes the Vale linting warnings and/or suggestions.

### DIFF
--- a/docs/backend/guides/how-to/manage-golang-with-asdf.md
+++ b/docs/backend/guides/how-to/manage-golang-with-asdf.md
@@ -13,7 +13,9 @@ Besides being the home row of the left hand in the qwerty layout it is also a to
 
 ## Why ASDF?
 
-To find out more on why asdf was chosen please read [ADR 0056 - Use asdf to manage golang versoins in development](https://github.com/transcom/mymove/blob/master/docs/adr/0056-use-asdf-to-manage-golang-versions-in-development.md)
+To find out more on why asdf was chosen please read [ADR 0056 - Use asdf to
+manage Golang versions in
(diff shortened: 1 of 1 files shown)


docs/backend/guides/how-to/manage-golang-with-asdf.md
```diff
@@ -46,7 +54,7 @@ Then run `asdf global golang 1.15.10`
 
 ### Reload your shell
 
-Reload your shell, open a new shell or source your rc file.
+Reload your shell, open a new shell or source your resource file.
```
ronaktruss, line 57: I've generally seen the terminology rc file or .rc file. Personally I would keep this as rc file and add resource in parenthesis. While resource is probably more accurate since you can source your bash_profile, I'd still advocate for rc. Fwiw it also doesn't look like there is a consensus on what rc actually stands for. I had "learned"  it was run command/control (can't recall which, but I see reference to both) in the context of unix but also see that resource is also a common definition.

rogeruiz: Looks like **Run Commands** would be a better name here. So we can keep them as `rc` file, or better yet name the files explicitly rather than have them be "rc files" since this would be an actual file that people edit, `bashrc`, `zshrc` or `fishrc`.

rogeruiz: Honestly though, on second thought I don't see a big value in keeping these docs around for long. The official `asdf` documentation is going to be kept more up-to-date than whatever we run.

rogeruiz:
Closing this as I am unable to trigger a `test_build` without modifying things and also #61 has not beed merged in. So I'm going to remove this branch and close the PR. We can revisit this work at another time after #61 has been successfully merged into `main`.

PR #105, opened 2022-01-21
[MB-10309] Add GH action for checking for Broken Links

rogeruiz:
Pairing with @donaldthai on this one. This is hopefully all we need to do to add Broken Link reporting to this repository.

### DIFF
--- a/.github/workflows/broken-links.yml
+++ b/.github/workflows/broken-links.yml
@@ -3,7 +3,7 @@ name: Using htmlproofer to report out all Broken Links
 on: [push]
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/broken-links.yml
```diff
@@ -0,0 +1,29 @@
+name: Using htmlproofer to report out all Broken Links
+
+on: [push]
+
+jobs:
+  prose:
```
rogeruiz, line 6: This needs to be changed to something like `check_broken_links` in order to be working properly.

PR #107, opened 2022-01-26
Fix the build

rogeruiz:
The last PR #106  that I merged in broke the build. This PR adds a new Action Workflow to build Docusaurus so that PRs that may break the build don't get merged in. I also refactor the configuration and the homepage to avoid the break in the build that #106 caused due to ES6 import statements.

### DIFF
--- a/.github/workflows/test-builds.yml
+++ b/.github/workflows/test-builds.yml
@@ -1,0 +1,19 @@
+name: Test build Docusaurus source
+
+on: [push]
+
+jobs:
+  test_build:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


YanZ777 (approved):

rogeruiz:
@YanZ777 This should now be ready to review. We should see the `test_build` action passing before we're allowed to merge PRs now. https://github.com/transcom/transcom-infrasec-com/pull/1766#issuecomment-1022275992

jacquelineIO (approved):
looks good. did walk through with Roger on this.

PR #108, opened 2022-01-26
MB-10593 Move locust docs from load testing repo to here

felipe-lee:
I've moved the existing load testing docs here instead of all being in the milmove load testing repo README.

I've split most of the docs into more digestible chunks, though there is likely a better flow. Let me know if there's particular things you'd like to see differently. 

This is PR that will need to be merged before https://github.com/transcom/milmove_load_testing/pull/115. 

Also, there were two things that came up during the load testing recording regarding the docs that I addressed as I moved the docs over:
* There are some doc updates that I realized I should make from putting that workflow together, namely info on `MilMoveData` and `DataType`. 
    * <http://localhost:4000/mymove-docs/docs/tools/locust/fake-data-generation#base-fake-data-generator>
* There is a section that talks about the typical steps to add a new load test. Something that came up during the recording was that it would be helpful to add links to the corresponding sections of the docs that talk more in depth about what each step entails. 
    *  <http://localhost:4000/mymove-docs/docs/tools/locust/adding-or-editing-load-tests>

## Setup

<details>
<summary>💻</summary>

* If you haven't set up this repo before, follow the setup instructions: https://github.com/transcom/mymove-docs#running-locally-on-macos
* Start docusaurus:

    ```shell
    yarn start
    ```

* Navigate to the new locust docs section: <http://localhost:4000/mymove-docs/docs/tools/locust/locust>

</details>


rogeruiz:
Thanks @felipe-lee. I am not going to check this now. but #107 is adding `test_build` to the PR review process. My only suggestion right now is to make sure that you run `yarn build && yarn serve` locally and make sure that the site works. Right now we don't have a way to prevent PRs that run successfully on `yarn start` but fail to `yarn build`. But we will once #107 and transcom/transcom-infrasec-com#1766 are merged in.

felipe-lee:
> Thanks @felipe-lee. I am not going to check this now. but #107 is adding `test_build` to the PR review process. My only suggestion right now is to make sure that you run `yarn build && yarn serve` locally and make sure that the site works. Right now we don't have a way to prevent PRs that run successfully on `yarn start` but fail to `yarn build`. But we will once #107 and [transcom/transcom-infrasec-com#1766](https://github.com/transcom/transcom-infrasec-com/pull/1766) are merged in.

Ah, thanks for the heads up, I'll try that tomorrow to make sure it works as expected.

rogeruiz:
@felipe-lee we should now have restrictions that mean that `test_build` must pass in order to merge in PRs. So you shouldn't have to do it manually. 

https://github.com/transcom/transcom-infrasec-com/pull/1766#issuecomment-1022275992

docs/tools/locust/api-parsers.md
```diff
@@ -0,0 +1,167 @@
+---
+sidebar_position: 11
+---
+# API Parsers
```
duncan-truss, line 4: I will say there has been confusion in the past between the Prance parser being able to parse the swagger file versus the Spectral linter finding it valid.  I don't know if that's the place to call this out but I think it's easy to think they are one in the same.

rogeruiz: Interesting @duncan-truss so you're saying that the Spectral linter and the Prance parser don't agree on the validity of our Swagger files? Do you know of an example where this happens? It makes sense to be explicit about this difference so folks aren't confused by the discrepancy.

rogeruiz:
@felipe-lee this, _for now_, won't prevent you from merging this in, but Vale is suggesting that you have some spelling and grammatical errors. Some of these can be addressed by capitalizing things. Other's can also be fixed by adding the terms to the `vale/Vocab/OSS-Docs/accept.txt` in order to resolve the suggestions.

[=> All the Vale suggestions and warnings in a single place for this PR](https://github.com/transcom/mymove-docs/pull/108/checks?check_run_id=4953321229)

docs/tools/locust/api-parsers.md
duncan-truss: Yea pre-commit is setup to lint the prime and support APIs according to the spectral linter with some custom rules thrown in https://github.com/transcom/mymove/blob/master/swagger-def/.spectral.yml.  The internal and ghc APIs are not up to the style guide enforced here for things like paths being kebab-case or body properties being camelCase because of there being legacy code that requires client side changes if the API shape changes.

Another pre-commit step to separately parse all of the APIs using Prance would actually be a nice way to know if we've broken the load test parser.  Load tests could be broken if not updated regardless if the API is parseble though for other reasons but it would be a start.

docs/tools/locust/working-with-load-tests.md
```diff
@@ -0,0 +1,28 @@
...
+
+One thing to note is that `locust` has a concept of `host` that differs a bit from how we use it. `locust` defaults to
+using `host` in the `User`/`TaskSet` context as a base domain to then add on to when you make requests, but because we
+have a need to make requests outside the `User`/`TaskSet` context, we've taken a slightly different approach.
+
+We use `locust`'s `host` value as a key to know which `mymove` server we want to target (mentioned in the 
+[Running Locust Locally](./running-locust-locally) section). We use `locust`'s `host` value combined with an `Enum`
+called `MilMoveEnv` to know what environment we're targeting. Then we combine it with a helper class called 
+`MilMoveRequestPreparer` which helps us build proper URLs and enables us to make requests more easily in and out of the 
+`User`/`TaskSet` context.
```
duncan-truss, line 23: This is helpful, I still get tripped up by the concept of where locust is running versus where milmove is running and what is a host vs environment.

docs/tools/locust/taskset.md
```diff
@@ -0,0 +1,308 @@
...
+
+Tasks are distinct functions, or callables, that tell Locust what to do during load testing. A task
+is, in essence, a load test. `TaskSet` classes are a way to link tasks together and keep the code
+organized. Task sets and functions should all be defined in python files in the `tasks/` directory.
+
+It is possible for a user class to have more than one task set, but it's important to keep in mind
+is that if a user has more than one task set, they will only ever switch between the task sets if
+you remember to have the task set stop at some point. Otherwise, the user will just stay stuck on
+their first task set until the load tests end. Examples will be included below.
+
```
duncan-truss, line 14: It may be useful to note that the tasksets get run in a randomized order and with the same proportion unless we override the weight per user class or task.

duncan-truss (approved):
This looks good to me! The last thing I might call out is that someone has to update the load test environment manually because master doesn't get deployed there automatically.  I don't know if that's captured elsewhere?

docs/tools/locust/api-parsers.md
felipe-lee: Ah good point. I can add a comment about this.

felipe-lee:
> @felipe-lee this, _for now_, won't prevent you from merging this in, but Vale is suggesting that you have some spelling and grammatical errors. Some of these can be addressed by capitalizing things. Other's can also be fixed by adding the terms to the `vale/Vocab/OSS-Docs/accept.txt` in order to resolve the suggestions.
> 
> [=> All the Vale suggestions and warnings in a single place for this PR](https://github.com/transcom/mymove-docs/pull/108/checks?check_run_id=4953321229)

Thanks for calling this out, let's see if the commit I just pushed fixes the issues.

docs/tools/locust/working-with-load-tests.md
felipe-lee: Yeah, I kind of wish we'd done it a bit differently now that I've been messing with it more. Then again, maybe having this called out is enough 🤷🏼

docs/tools/locust/taskset.md
felipe-lee: Ah good point, I'll add a note about that.

felipe-lee:
> This looks good to me! The last thing I might call out is that someone has to update the load test environment manually because master doesn't get deployed there automatically. I don't know if that's captured elsewhere?

@duncan-truss , good question. It's mentioned on this confluence page: https://dp3.atlassian.net/wiki/spaces/MT/pages/1512603655/Deploying+load-tests+to+AWS

I mention the confluence docs and just added a note saying deployment docs are there. Do you think that's enough?

PR #110, opened 2022-01-27
Update nix setup

gidjin:
The current nix configuration was failing on my M1 macbook. This PR updates the configuration to work for the latest version of nix and M1 setup.

Should be reviewed by someone using nix.

The following should pass and start docusaurus.

```sh
direnv allow
./nix/update.sh
yarn install
yarn start
```

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -2,7 +2,7 @@
 if [ ! -r .nix-disable  ] && has nix-env; then
   # set NIX_PROFILE so nix-env operations don't need to manually
   # specify the profile path
-  export NIX_PROFILE=/nix/var/nix/profiles/mymove-docs
+  export NIX_PROFILE="/nix/var/nix/profiles/per-user/${LOGNAME}/mymove-docs"
 
   # Having NIX_SSL_CERT_FILE set means go won't use macOS keychain
   # based certs
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -6,18 +6,18 @@ in buildEnv {
   paths = [
     (import (builtins.fetchGit {
       # Descriptive name to make the store path easier to identify
-      name = "nodejs-14.17.1";
+      name = "nodejs-14.18.3";
       url = "https://github.com/NixOS/nixpkgs/";
       ref = "refs/heads/nixpkgs-unstable";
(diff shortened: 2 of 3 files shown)


.envrc
```diff
@@ -2,7 +2,7 @@
 if [ ! -r .nix-disable  ] && has nix-env; then
   # set NIX_PROFILE so nix-env operations don't need to manually
   # specify the profile path
-  export NIX_PROFILE=/nix/var/nix/profiles/mymove-docs
+  export NIX_PROFILE="/nix/var/nix/profiles/per-user/${LOGNAME}/mymove-docs"
```
felipe-lee, line 5: This works for me, but I wonder if we should have a check for whether or not the person has write permissions to the profile like we do in the mymove repo: 

https://github.com/transcom/mymove/blob/3661c4f0d2785d06307815374e020e109364f3df/.envrc#L302-L306

felipe-lee (approved):
I don't have an M1, but as a nix user, this works for me. I just have one comment, but I think it's fine either way probably.

.envrc
rogeruiz: I'm with @felipe-lee on this. Let's standardize across the board. I never considered this project o need something like Nix, though I'm okay with it being included. But if we're gonna Nix let's Nix the way we do across the board. I'd strongly recommend we modify this to check and have error handling. 

```suggestion
  if [ -w "/nix/var/nix/profiles/per-user/${LOGNAME}" ]; then
    export NIX_PROFILE="/nix/var/nix/profiles/per-user/${LOGNAME}/mymove-docs"
  else
    log_error "Cannot determine writable location for your NIX_PROFILE"
  fi

```

gidjin: I would be tempted to add it in but for the thread in the EngPlaybook where Drew was thinking it was clearer to allow the system to throw the error https://github.com/trussworks/Engineering-Playbook/pull/210#discussion_r767888609

rogeruiz: does the system through the error properly? do we know what the error is? because this is a pre-message before Nix even runs versus what we're saying about having Nix throwing errors are different topics. But I'm not going to prevent this from making it into the repo. 👍🏽

felipe-lee: Ah ok, sounds good

PR #111, opened 2022-02-17
Add Mermaid diagram support

rogeruiz:
This PR adds Mermaid diagram support using code blocks. Similar to GitHub, we're able to add them as code blocks.

### This code block

<pre><code>
```mermaid
flowchart LR
   a-->b
   b--->c
   b-->a
```
</code></pre>

### Produces this

```mermaid
flowchart LR
   a-->b
   b--->c
   b-->a
```

### DIFF
--- a/docs/tools/locust/running-load-tests.md
+++ b/docs/tools/locust/running-load-tests.md
@@ -3,15 +3,37 @@ sidebar_position: 2
 ---
 # Running Load Tests
 
-There are several ways you can run load tests. You can run `locust` locally or use the deployed version (on AWS). You 
+There are several ways you can run load tests. You can run `locust` locally or use the deployed version (on AWS). You
 can also use a local `mymove` server or the one deployed in the load test environment (on `loadtest.dp3.us`). One thing
-to note is that you can't point the deployed `locust` instance at a local `mymove` server. This means you have three 
-possible combinations:
+to note is that you can't point the deployed `locust` instance at a local `mymove` server.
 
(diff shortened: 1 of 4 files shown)


rogeruiz:
[read more about diagram syntax here](https://mermaid-js.github.io/mermaid/#/n00b-syntaxReference)

docs/tools/locust/running-load-tests.md
```diff
@@ -3,15 +3,37 @@ sidebar_position: 2
...
+flowchart RL
+    subgraph Environments
+        subgraph Local
+            direction TB
+            l_customer[(Customer App)]
+            l_office[(Office App)]
+            l_prime[(Prime API)]
+        end
+
+        subgraph AWS
```
felipe-lee, line 21: Do you think it'd be good to include the `loadtest.dp3.us` in here? To differentiate between this AWS and the AWS locust instance.

felipe-lee (approved):
LGTM! I just left one comment, but I'm fine with it either way.

docs/tools/locust/running-load-tests.md
rogeruiz: I figured the differentiation would be that the AWS locust is outside of the "Environments > AWS". The reason I went with a short name here is because it's used in the diagram. Also, I didn't want to tie the name to an actual URL. It's possible our URL could change. And it's less likely that our IaaS would change. But it's not a terrible idea if you think it should read the URL. I'm okay with it either way.

PR #113, opened 2022-02-18
Add in basic Danger support

rogeruiz:


rogeruiz:
Doesn't look like the Danger JS action supports plugins, so I'm refactoring this to use Yarn instead.

rogeruiz:
Coverting back to draft as I still have permission issues for the Danger JS workflow action.

PR #117, opened 2022-03-10
[ MB-11902 ] Add in the other APIs

rogeruiz:
This was suggested by @mboren for ease of use. I agreed and used my on-call duties to help facilitate this. These docs are already public so I do not see the harm in adding them in to our public documentation. But please feel free to review these changes locally.


jenniferlynparsons (approved):
LGTM!

PR #119, opened 2022-03-11
[ MB-11957 ] update migration docs

NamibiaTorres:
This PR is associated with JIRA ticket [MB-11957](https://dp3.atlassian.net/browse/MB-11957)

Review instruction:
* look at the delete migrations sections and confirm that the instructions are correct.

### DIFF
--- a/docs/backend/setup/database-migrations.md
+++ b/docs/backend/setup/database-migrations.md
@@ -10,7 +10,7 @@ If you need to change the database schema, you'll need to write a migration. The
 2. [Add the new SQL to the generated file](#writing-migrations)
 3. [Set up your database](#Setup)
 4. [Run the migrations](#Running-Migrations)
-5. [Delete a migration](#Delete-Migration)
+5. [Delete a migration](#Delete-Migrations)
 6. Test your new migration
 
 After your testing, if you find that you need to change your migration, you'll need to reset your DB (`make db_<env>_reset`) and rerun the migrations to make sure your updates are reflected in the local DB instance. 
--- a/docs/backend/setup/database-migrations.md
+++ b/docs/backend/setup/database-migrations.md
@@ -7,10 +7,11 @@ sidebar_position: 4
 If you need to change the database schema, you'll need to write a migration. These are the general steps you'll need to follow:
 
 1. [Generate a new migration file](#Creating-Migrations)
-1. [Add the new SQL to the generated file](#writing-migrations)
-1. [Set up your database](#Setup)
-1. [Run the migrations](#Running-Migrations)
-1. Test your new migration
+2. [Add the new SQL to the generated file](#writing-migrations)
+3. [Set up your database](#Setup)
+4. [Run the migrations](#Running-Migrations)
+5. [Delete a migration](#Delete-Migration)
+6. Test your new migration
 
 After your testing, if you find that you need to change your migration, you'll need to reset your DB (`make db_<env>_reset`) and rerun the migrations to make sure your updates are reflected in the local DB instance. 
 
@@ -157,6 +158,15 @@ Migrations are run by the `milmove migrate` command. This allows us to leverage 
 - `make db_deployed_migrations_migrate`
 
 The reason to use a `make` target is because it will correctly set the migration flag variables and target the correct database with environment variables.
+
+## Delete Migrations
+In the event that you need to make an edit to a migration that you have just created prior to merging it into the main branch, 
+you can delete the migration and recreate it after making your edits. To do this, follow these steps:
+
+1. Delete the migration file in the manifest 
+2. Rerun the migrations using: `make db_dev_reset db_dev migrate` 
+3. Recreate the migration file in the manifest 
+4. Rerun the migration
 
 ## Secure Migrations
 
--- a/docs/backend/setup/database-migrations.md
+++ b/docs/backend/setup/database-migrations.md
@@ -161,7 +161,11 @@ The reason to use a `make` target is because it will correctly set the migration
 
 ## Update Migrations Locally
 In the event that you need to make an edit to a migration that you have just created prior to merging it into the main branch, 
-you can update the migration with your edits and rerun it using: `make db_dev_reset db_dev_migrate` 
+you can update the migration with your edits and rerun it using: 
+
+```sh
+make db_dev_reset db_dev_migrate
+``` 
 
 This command will reset the database and re-add all migrations, including your updated one. Double check that you see 
 the database changes after making your edits and running the `make` commands. 
--- a/docs/backend/setup/database-migrations.md
+++ b/docs/backend/setup/database-migrations.md
@@ -10,7 +10,7 @@ If you need to change the database schema, you'll need to write a migration. The
 1. [Add the new SQL to the generated file](#writing-migrations)
 1. [Set up your database](#Setup)
 1. [Run the migrations](#Running-Migrations)
-1. [Rollback migrations locally](#Rollback-Migrations-Locally)
+1. [Update migrations locally](#Update-Migrations-Locally)
 1. Test your new migration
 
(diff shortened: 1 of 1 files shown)


docs/backend/setup/database-migrations.md
```diff
@@ -158,6 +159,15 @@ Migrations are run by the `milmove migrate` command. This allows us to leverage
 
 The reason to use a `make` target is because it will correctly set the migration flag variables and target the correct database with environment variables.
 
+## Delete Migrations
```
felipe-lee, line 162: Wait so is this saying to delete the migration file itself, or only remove the line in the manifest that points to it?

Mostly because if it's the latter, then to me it's not so much about "deleting" a migration, as making edits and re-running a migration, in which case running the `make db_dev_reset db_dev migrate` command would handle it.

NamibiaTorres: Delete the migration file itself. Although if there is a better way to delete a migration, I'd rather offer that option...

gidjin: I think I failed to explain this well when we paired. You only need to remove the file from the manifest, you don't need to delete the migration file. If you do delete the file the build will complain of a missing migration since it would still be in the manifest.

I would also add a note that this is only if you want to undo a migration you've added locally. For the deployed environments the process is to create a new migration that un-does the change.

Let me know if this helps if not I'm happy to connect on Monday :)

NamibiaTorres: That's what I meant. The migration file that is in the manifest, not the entire folder. Does that clear things up or is there another migration file that I didn't understand?

felipe-lee: I think maybe in that case this should be split into two sections. Like one for running a changed migration (that hasn't made it to a deployed env). Which I think would be two steps:

1. Make changes to migration
2. Run

    ```shell
    make db_dev_reset db_dev migrate
    ```

And a section on deleting a migration fully. This one could have two things, one about it being local only still, so 3 steps:

1. Remove the line in the manifest pointing at the migration file.
2. Delete the migration file
3. Run 

    ```shell
    make db_dev_reset db_dev migrate
    ```

And a part mentioning that if doing it for a migration that has already been run against the deployed environments, that we need to follow our zero-downtime migrations practice (which is on this page).

docs/backend/setup/database-migrations.md
```diff
@@ -7,10 +7,11 @@ sidebar_position: 4
...
 1. [Generate a new migration file](#Creating-Migrations)
-1. [Add the new SQL to the generated file](#writing-migrations)
-1. [Set up your database](#Setup)
-1. [Run the migrations](#Running-Migrations)
-1. Test your new migration
+2. [Add the new SQL to the generated file](#writing-migrations)
+3. [Set up your database](#Setup)
+4. [Run the migrations](#Running-Migrations)
+5. [Delete a migration](#Delete-Migrations)
+6. Test your new migration
```
gidjin, line 14: I believe that docusaurus automatically adds the right numbers when it renders the real page so these should all be `1.` still.

docs/backend/setup/database-migrations.md
gidjin: Do we ever delete migrations once they have been deployed? Not sure that is something we need to document, at least until we actually need to do so.

docs/backend/setup/database-migrations.md
NamibiaTorres: Ok, thanks!

docs/backend/setup/database-migrations.md
```diff
@@ -7,10 +7,11 @@ sidebar_position: 4
...
 1. [Generate a new migration file](#Creating-Migrations)
-1. [Add the new SQL to the generated file](#writing-migrations)
-1. [Set up your database](#Setup)
-1. [Run the migrations](#Running-Migrations)
-1. Test your new migration
+2. [Add the new SQL to the generated file](#writing-migrations)
+3. [Set up your database](#Setup)
+4. [Run the migrations](#Running-Migrations)
+5. [Rollback migrations locally](#Rollback-Migrations-Locally)
+6. Test your new migration
```
rogeruiz, line 14: Let's have Markdown create the numbered list for us rather than manually having to maintain it. 


```suggestion
1. [Add the new SQL to the generated file](#writing-migrations)
1. [Set up your database](#Setup)
1. [Run the migrations](#Running-Migrations)
1. [Rollback migrations locally](#Rollback-Migrations-Locally)
1. Test your new migration
```

felipe-lee (approved):
LGTM, I made a minor suggestion, but it's not vital, so up to you whether it goes in or not.

docs/backend/setup/database-migrations.md
```diff
@@ -158,6 +159,13 @@ Migrations are run by the `milmove migrate` command. This allows us to leverage
 
 The reason to use a `make` target is because it will correctly set the migration flag variables and target the correct database with environment variables.
 
+## Update Migrations Locally
+In the event that you need to make an edit to a migration that you have just created prior to merging it into the main branch, 
+you can update the migration with your edits and rerun it using: `make db_dev_reset db_dev_migrate`
```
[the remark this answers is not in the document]
NamibiaTorres: > Minor suggestion, this makes it so there's a "copy" button on the right so people can quickly copy the command. Kind of like the first command in [the Creating Secure Migration section](https://transcom.github.io/mymove-docs/docs/backend/setup/database-migrations/#creating-secure-migrations). Contrasted with the commands in [the Running Migrations section](https://transcom.github.io/mymove-docs/docs/backend/setup/database-migrations/#running-migrations). See how in the latter, they're formatted, but have to be highlighted to be copied, while in the former, there's a copy button on the right when you hover over the box.

Thanks @felipe-lee , do you happen to understand the difference between writing shell and writing console? I see we do both in this file.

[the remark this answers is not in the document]
felipe-lee: I'm not sure what the difference is tbh. I tried looking into it and Docusaurus has this section on supported languages:
 https://docusaurus.io/docs/markdown-features/code-blocks#supported-languages

Based on that, it looks like the default supported languages are these:
https://github.com/FormidableLabs/prism-react-renderer/blob/master/src/vendor/prism/includeLangs.js

But neither `shell` nor `console` are on there...You can extend the default list, but looking at our docusaurus config, it looks like we haven't added more languages: https://github.com/transcom/mymove-docs/blob/071f511b101d71df7aeabe893fb1a7a1191c946c/docusaurus.config.js#L149-L152

Digging into it some more, it looks like this is the full list of supported languages: https://prismjs.com/#supported-languages

That docs says `shell` is an alias for `bash` so I guess that's why that one works, but there's no mention of `console`

I tried looking into it some more, and ended up on [a thread with people talking about the syntax highlighters github](https://stackoverflow.com/questions/20303826/highlight-bash-shell-code-in-markdown-files) uses (or people think they use?). That pointed me to these two different highlighters:

* linguist
    * `console`: https://github.com/github/linguist/blob/3c3b037910006fc2f1a9bb34b2c4e9cde062206c/lib/linguist/languages.yml#L5837-L5848
    * `shell`: https://github.com/github/linguist/blob/3c3b037910006fc2f1a9bb34b2c4e9cde062206c/lib/linguist/languages.yml#L5753-L5824
    * They seem to be mostly the same except `shell` defines file names and one of the attributes, but I don't know enough about this to know what that difference means.
* highlight.js
    * https://github.com/highlightjs/highlight.js/blob/main/SUPPORTED_LANGUAGES.md
    * This one treats `shell` and `console` the same way it seems.

So yeah, I'm not entirely sure. Based on the thread I linked above where people were talking about the gh syntax one, it seems `console` might be able to handle the terminal prompt, but I haven't really tested it. I personally lean toward `shell` since it seems to be a common default language that syntax highlighters support, but I could be wrong. Maybe others know more.

PR #120, opened 2022-03-17
MB-11996 update migration docs

felipe-lee:
[JIRA ticket](https://dp3.atlassian.net/browse/MB-11996)

This adds/updates information about:

* pluralization of sql table name
* pointers vs not for optional/required fields
* `TableName` function for names that are different from the default generated name
* Adding comments to fields

Also fixed a few links on the page.

### DIFF
--- a/docs/backend/setup/database-migrations.md
+++ b/docs/backend/setup/database-migrations.md
@@ -111,7 +111,7 @@ See the [conventions Pop follows](https://www.gobuffalo.io/en/docs/db/getting-st
 :::
 
 1. [Generate a new migration file](#creating-migrations).
-2. Open the generated SQL file and add the logic to create your table.  When creating the SQL you may write the migration like this:
+1. Open the generated SQL file and add the logic to create your table.  When creating the SQL you may write the migration like this:
 
     ```sql
     CREATE TABLE cats
@@ -138,12 +138,12 @@ See the [conventions Pop follows](https://www.gobuffalo.io/en/docs/db/getting-st
 
     1. The SQL keywords don't need to be capitalized, but it can be nice to differentiate some of them from certain 
        things like the field names or types.
-    2. We want to add columns for every column other than the `id`, `created_at`, and `updated_at` fields. This makes it
-       easier for future folks to understand what the columns were meant for.
-    3. You might see some migrations use `varchar` instead of `text`. It's the same thing in the end, see
+    1. We want to add comments for every column other than the `id`, `created_at`, and `updated_at` fields. This makes 
+       it easier for future folks to understand what the columns were meant for.
+    1. You might see some migrations use `varchar` instead of `text`. It's the same thing in the end, see
        [Postgresql Character Types](https://www.postgresqltutorial.com/postgresql-char-varchar-text/) for more info.
     
-3. Now you can create a new file in `pkg/models/` named after your new model. So for this, we'll have a new file 
+1. Now you can create a new file in `pkg/models/` named after your new model. So for this, we'll have a new file 
     `pkg/models/cat.go` that will look something like this:
 
     ```go
@@ -175,25 +175,26 @@ See the [conventions Pop follows](https://www.gobuffalo.io/en/docs/db/getting-st
 
    Some things to note:
     1. The field names here are pascal case, while in the struct tags, we use the snake case version. 
-    2. Notice how we used `string` for `Name`, while we used `*string` for `Bio`. We have a loose convention around
+    1. Notice how we used `string` for `Name`, while we used `*string` for `Bio`. We have a loose convention around
        using pointers for optional (nullable) fields, and the concrete type for required fields.
         1. One important thing to note here is that this will affect patch requests if you want the API caller to be 
            able to make a request without passing in all the non-nullable fields each time. Some options which we use 
            for existing models:
             1. Make the field a pointer on the model. This means you will need to have validation somewhere though to
                ensure that we save a valid value to the DB in the end.
-            2. Follow a pattern similar to the
+            1. Follow a pattern similar to the
                [PPM shipment updater](https://github.com/transcom/mymove/blob/088b87605f19b18b022ff718fcbd5eb8f6962585/pkg/services/ppmshipment/ppm_shipment_updater.go#L44-L49)
                which merges the new and old shipment _before_ validating the shipment.
-    3. We define the plural `type` under the main struct as a slice of the type.
-       
-
+    1. We define the plural `type` under the main struct as a slice of the type.
+       1. You might see some models that don't follow this pattern, but generally we prefer to follow it so if you are
+          adding a new table it's best to stick with this pattern.
+    
 1. Now you will want to run the migration to test it out with `make db_dev_migrate`.
 
 #### Model/Table Names With Acronyms
 
 If your model/table name has an acronym in it, e.g. `PPMShipment`/`ppm_shipments`, you will need to define a receiver 
-function called `TableName` for the struct that helps `Pop` know what the table name should be. An example is our 
+function called `TableName` for the model struct that helps `Pop` know what the table name should be. An example is our 
 [`PPMShipment` model's receiver function](https://github.com/transcom/mymove/blob/9c2a281dbd777b77064c1ae563531a3f0c7bf9d0/pkg/models/ppm_shipment.go#L62-L65):
 
 ```go
--- a/docs/backend/setup/database-migrations.md
+++ b/docs/backend/setup/database-migrations.md
@@ -182,6 +182,21 @@ See the [conventions Pop follows](https://www.gobuffalo.io/en/docs/db/getting-st
 
 1. Now you will want to run the migration to test it out with `make db_dev_migrate`.
 
+#### Model/Table Names With Acronyms
+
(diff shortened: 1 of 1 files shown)


docs/backend/setup/database-migrations.md
```diff
@@ -98,56 +98,112 @@ When you are writing the code for your new migration, there are a few things you
...
-  Column1   string    `json:"column1" db:"column1"`
-  Column2   *string   `json:"column2" db:"colunn2"`
-  CreatedAt time.Time `json:"created_at" db:"created_at"`
-  UpdatedAt time.Time `json:"updated_at" db:"updated_at"`
-}
+    Some things to note:
+
+    1. The SQL keywords don't need to be capitalized, but it can be nice to differentiate some of them from certain 
+       things like the field names or types.
+    2. We want to add columns for every column other than the `id`, `created_at`, and `updated_at` fields. This makes it
```
gidjin, line 141: I think you meant `We want to add comments for every column...` here. Using `1.` for number lists is the convention as I understand since docusaurus will auto number them anyway.

docs/backend/setup/database-migrations.md
```diff
@@ -98,56 +98,112 @@ When you are writing the code for your new migration, there are a few things you
...
+       using pointers for optional (nullable) fields, and the concrete type for required fields.
+        1. One important thing to note here is that this will affect patch requests if you want the API caller to be 
+           able to make a request without passing in all the non-nullable fields each time. Some options which we use 
+           for existing models:
+            1. Make the field a pointer on the model. This means you will need to have validation somewhere though to
+               ensure that we save a valid value to the DB in the end.
+            2. Follow a pattern similar to the
+               [PPM shipment updater](https://github.com/transcom/mymove/blob/088b87605f19b18b022ff718fcbd5eb8f6962585/pkg/services/ppmshipment/ppm_shipment_updater.go#L44-L49)
+               which merges the new and old shipment _before_ validating the shipment.
+    3. We define the plural `type` under the main struct as a slice of the type.
```
gidjin, line 188: I like this convention, though I know not all our models follow this pattern. I wonder if we should note that or not. Maybe it's enough to say that it should be there.

docs/backend/setup/database-migrations.md
```diff
@@ -98,56 +98,112 @@ When you are writing the code for your new migration, there are a few things you
...
+       
+
+1. Now you will want to run the migration to test it out with `make db_dev_migrate`.
```
gidjin, line 191: Not sure why but this is rendering as a `1.` when I start docusaurus. It's minor nit though so feel free to ignore.

docs/backend/setup/database-migrations.md
```diff
@@ -98,56 +98,112 @@ When you are writing the code for your new migration, there are a few things you
...
+
+#### Model/Table Names With Acronyms
+
+If your model/table name has an acronym in it, e.g. `PPMShipment`/`ppm_shipments`, you will need to define a receiver 
+function called `TableName` for the struct that helps `Pop` know what the table name should be. An example is our
```
gidjin, line 196: Maybe add model next to struct in this sentence?

> ... you will need to define a receiver function called `TableName` for the model struct that helps ...

gidjin (commented):
This looks good, found one typo that I think needs to be fixed and a few nits you are free to ignore :) Thank you for updating the docs!

docs/backend/setup/database-migrations.md
felipe-lee: 🤦🏼 thanks fro catching that typo

Ugh, yeah I keep having to fight my editor because it's trying to be helpful by auto-updating the numbers for me whenever I add one and I forgot to got back at the end to fix them.

docs/backend/setup/database-migrations.md
felipe-lee: Ah, I think because of the two blank lines, should only be one.

docs/backend/setup/database-migrations.md
felipe-lee: I didn't realize we had some without that. I think it's good to call out because I'm sure others will come across the different ones and wonder which to follow.

docs/backend/setup/database-migrations.md
```diff
@@ -98,56 +98,113 @@ When you are writing the code for your new migration, there are a few things you
...
-  Column1   string    `json:"column1" db:"column1"`
-  Column2   *string   `json:"column2" db:"colunn2"`
-  CreatedAt time.Time `json:"created_at" db:"created_at"`
-  UpdatedAt time.Time `json:"updated_at" db:"updated_at"`
-}
+    Some things to note:
+
+    1. The SQL keywords don't need to be capitalized, but it can be nice to differentiate some of them from certain 
+       things like the field names or types.
+    1. We want to add comments for every column other than the `id`, `created_at`, and `updated_at` fields. This makes
```
pearl-truss, line 141: Nit: might want to call out that a comment was also added for the new table

docs/backend/setup/database-migrations.md
```diff
@@ -98,56 +98,113 @@ When you are writing the code for your new migration, there are a few things you
...
+   Some things to note:
+    1. The field names here are pascal case, while in the struct tags, we use the snake case version. 
+    1. Notice how we used `string` for `Name`, while we used `*string` for `Bio`. We have a loose convention around
+       using pointers for optional (nullable) fields, and the concrete type for required fields.
+        1. One important thing to note here is that this will affect patch requests if you want the API caller to be 
+           able to make a request without passing in all the non-nullable fields each time. Some options which we use 
+           for existing models:
+            1. Make the field a pointer on the model. This means you will need to have validation somewhere though to
+               ensure that we save a valid value to the DB in the end.
+            1. Follow a pattern similar to the
```
pearl-truss, line 185: It's a great that you added a concrete code example here 🎉

pearl-truss (approved):

docs/backend/setup/database-migrations.md
```diff
@@ -98,56 +98,113 @@ When you are writing the code for your new migration, there are a few things you
...
+        UpdatedAt  time.Time   `json:"updated_at" db:"updated_at"`
+        Name       string      `json:"name" db:"name"`
+        Birthday   *time.Time  `json:"birthday" db:"birthday"`
+        GotchaDay  *time.Time  `json:"gotcha_day" db:"gotcha_day"`
+        Bio        *string     `json:"bio" db:"bio"`
+        Weight     *unit.Pound `json:"weight" db:"weight"`
+    }
+    
+    // Cats is a list of Cats
+    type Cats []Cat
```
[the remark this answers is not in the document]
felipe-lee: To be honest, I don't and haven't had to really work with them. Do you have an example in particular? I'm thinking that at least for when it's a return type, declaring its name in the function definition could make it easier to work with since you could append to the return var directly without having to initialize it yourself. I guess for parameters I wonder if there's a need to declare the variable being worked with as a slice of the model directly instead of using the plural type, which is why I assume there's a need for type conversions.   

cc: @gidjin, you mentioned you liked this pattern, not sure if you know more pros/cons.

docs/backend/setup/database-migrations.md
gidjin: I'm only good at catching typos because I've made so many. Sorry your editor is fighting you. I wonder if we could add a linter or something that would help fix it on save or something.

docs/backend/setup/database-migrations.md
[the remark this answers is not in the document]
gidjin: If I gave the impression I liked the pattern sorry for that. I'm indifferent too it tbh. My desire is for us to be consistent so either we have this pattern everywhere or we don't. As it stands we have it in some places but not others. So for this doc I'd like it to land on the recommendation so we have a place to point and steer PRs to consistency. 

I don't really see value of `Cats` over `[]Cat` both tell me it's a list (slice in go here I believe) of `Cat` models, and I can see an argument that the latter is more explicit about it.

docs/backend/setup/database-migrations.md
gidjin: Contradicting my self. I should clarify I like that we call out the pattern, I'm indifferent to the pattern itself. 🤦  sorry for the confusion

gidjin (approved):
LGTM, on the `type Cats []Cat` I can go either way, see comments elsewhere.

docs/backend/setup/database-migrations.md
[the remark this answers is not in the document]
gidjin: I will add that saying we should have plural types is easier than saying we shouldn't. Since the former means we just have to add the type if it's missing and nothing breaks. But the latter will require replacing all the `Cats` with `[]Cat` which isn't hard, just might be a lot of places. And one could argue that when going the other direction you should make the same search and replace, but if you don't the code still works. Hopefully this has been clear.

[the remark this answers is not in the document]
duncan-truss: I'm okay with us being consistent and keeping with the types because they are used in all of the has_many associations. My main point was just that this has tripped me up a bunch and I haven't come across a situation where this is a clear win.  I don't see Pop recommending them or anything https://www.gobuffalo.io/en/docs/db/relations/onetomany/#example

PR #121, opened 2022-03-24
[ MB-11958 ] Add some additional information about transactions

rogeruiz:
This was brought up in https://github.com/transcom/mymove/pull/8304#discussion_r833411905. I'm placing it in the documentation as part of the work for MB-11958 so that we can keep track why this refactor is being done.


michaelvaldes-truss (approved):
lgtm!

PR #122, opened 2022-03-28
Add guide for debugging with vscode

michaelvaldes-truss:


rogeruiz (approved):
lgtm 🌈

Thanks for writing this up. I wonder if once we have more documentation on Debugging, if we can refactor this for all the editors that Trussel's use on MilMove. But this is great start for now.

PR #124, opened 2022-03-29
[MB-11991] Update Go upgrade docs

reggieriser:
This updates the list of files that need to be changed as part of a Go upgrade.  This doc has gotten a little out of sync with the current state.

### DIFF
--- a/docs/backend/guides/how-to/upgrade-go-version.md
+++ b/docs/backend/guides/how-to/upgrade-go-version.md
@@ -22,10 +22,11 @@ Upgrading the Go version that we use happens in roughly these steps:
 
 ## 2. Upgrade Local Go Version
 
+### asdf
+
 For more details see [Manage golang with asdf](manage-golang-with-asdf.md)
 
 - Update `.tool-versions` to point to new version of golang
-  - If you've done some PATH sorcery to point to a specific Go version (as detailed [here](https://github.com/transcom/mymove#setup-prerequisites)), you'll have to update that as well
 - Run `asdf install` to install the new version (if you do not yet have it installed, see [How to get up and running](https://github.com/transcom/mymove/wiki/Manage-golang-with-asdf#how-to-get-up-and-running))
 - Run `asdf global golang <version>` to update global version as well
 - `go version` and `asdf which go` to check it worked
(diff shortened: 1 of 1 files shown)


docs/backend/guides/how-to/upgrade-go-version.md
```diff
@@ -58,20 +58,20 @@ go version go1.16.4 darwin/amd64
     - `scripts/gen-assets`
     - `scripts/gen-server`
     - `scripts/pre-commit-swagger-validate`
-    - `GOVERSION` in `scripts/prereqs`
+    - `scripts/run-e2e-test-docker`
     - `Makefile` the `docker_circleci` task
     - `nix/default.nix` (you will also need to [update the hash](https://ahobson.github.io/nix-package-search))
+    - .go-version
```
felipe-lee, line 64: I don't know if you also want to include the README since there's that example path for the GOROOT.

felipe-lee, line 64: I also wonder if the `nix/default.nix` bit should be moved to be in the previous section where the `.tool-versions` is since a `nix` user would do that instead to update their local `go` version.

docs/backend/guides/how-to/upgrade-go-version.md
```diff
@@ -58,20 +58,20 @@ go version go1.16.4 darwin/amd64
...
     - `go-version` in `.github/workflows/go-auto-approve.yml`
-    - `VERSION_NUMBER` in `scripts/check-go-version`
     - `go.mod`
 - Rerun the Go formatter on the codebase with `pre-commit run --all-files golangci-lint`
 - Regenerate mocks (in case the signatures have changed that we're mocking): `make mocks_generate`
 - Run `make e2e_test_docker` to test that the `Dockerfile.*local` files work  with the new image.
 - Commit the above changes and any reformatted code and make sure everything builds correctly on CircleCI
+- You may also want to check for an upgrade to the `golangci-lint` version, particularly if doing a major Go version upgrade
 
 ## 4. Notify Folks
```
felipe-lee, line 77: I wonder if linking to a previous post would be good in here so people have something to copy from or use as a reference.

felipe-lee (approved):
Left a few comments, but otherwise LGTM.

reggieriser:
@felipe-lee Thanks for reviewing!  I know you already approved, but I made changes based on each of your comments.  Could you do a quick re-review and make sure this seems to address appropriately?

felipe-lee (approved):
LGTM, thanks for making these updates!

PR #126, opened 2022-04-07
Fix syntax error/typo for soft-delete.md

pearl-truss:
Fix the syntax in the code block in the `DeleteExampleModel` example


felipe-lee (approved):
Awesome, thanks for fixing this

PR #127, opened 2022-04-15
Update upgrade-go-version.md

Ryan-Koch:
Adds a blurb to the doc with a template notification message I got from @reggieriser.


reggieriser (approved):
LGTM!  Thanks for adding this.  One nit suggestion is that you could use a blockquote to set off the template a little more from the surrounding text.

PR #129, opened 2022-04-26
Add ExcludeDeletedScope info and example to soft delete docs

duncan-truss:
Replaced the example query filtering out deleted records with a where clause to use the newer `Scope(utilities.ExcludeDeletedScope())` convenience call.

[Corresponding PR](https://github.com/transcom/mymove/pull/8445) that introduced the ExcludeDeletedScope

To test run:
1. `yarn install` && `yarn start`
2. Visit http://localhost:4000/mymove-docs/docs/backend/guides/how-to/soft-delete

Screen Shot 2022-04-26 at 3 58 27 PM

### DIFF
--- a/docs/backend/guides/how-to/soft-delete.md
+++ b/docs/backend/guides/how-to/soft-delete.md
@@ -26,15 +26,32 @@ type ExampleModel struct {
 
 If this has not been done, one must [create a migration](migrate-the-database.md) to make these changes.
 
-Furthermore, any queries to fetch the model must exclude those that have been 'soft deleted'.
+## Querying for non-deleted records
+Records that have been soft deleted will still exist in the database so we must filter them out in our database queries if we want to omit deleted data.
+
+The Pop ORM has a chainable method called [Scope](https://gobuffalo.io/documentation/database/scoping/) that we can use to append the where clause(s) to only include non-deleted records.
 
 ```go
-func FetchExampleModel(ctx context.Context, db *pop.Connection, session * auth.Session, id uuid.UUID) (ExampleModel, error) {
-    var exampleModel ExampleModel
-    err := db.Q().Where("example_models.deleted_at is null").Eager().Find(&exampleModel, id)
-    ...
+func FindShipment(ctx context.Context, shipmentID uuid.UUID) {
+	var shipments models.MTOShipments
+	ctx.DB().Scope(utilities.ExcludeDeletedScope()).All(&shipments)
 }
 ```
+
+By passing in the model we can also derive the table name if needed to disambiguate from other tables.
+
+```go
+func FindDocumentsWithUploads(ctx context.Context, uploaderID uuid.UUID) {
+	var documents models.Documents
+	ctx.DB().Scope(utilities.ExcludeDeletedScope(models.Document{}, models.UserUpload{})).
+		Join("user_uploads", "user_uploads.document_id = documents.id").
+		All(&documents)
+}
+```
+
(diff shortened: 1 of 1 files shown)


docs/backend/guides/how-to/soft-delete.md
```diff
@@ -26,16 +26,33 @@ type ExampleModel struct {
...
+```go
+func FindDocumentsWithUploads(ctx context.Context, uploaderID uuid.UUID) {
+	var documents models.Documents
+	ctx.DB().Scope(utilities.ExcludeDeletedScope(models.Document{}, models.UserUpload{})).
+		Join("user_uploads", "user_uploads.document_id = documents.id").
+		All(&documents)
+}
+```
+
+Unfortunately this will not filter any eager loaded associations so you may need to continue doing that depending on your query.  You should fallback to using a normal where clause if using an alias in your query or writing a RawQuery.
```
ronaktruss, line 52: Its probably worth including an example of when this query won't work and including an example of the fallback option.

docs/backend/guides/how-to/soft-delete.md
```diff
@@ -26,16 +26,33 @@ type ExampleModel struct {
...
-    var exampleModel ExampleModel
-    err := db.Q().Where("example_models.deleted_at is null").Eager().Find(&exampleModel, id)
-    ...
+func FindShipment(ctx context.Context, shipmentID uuid.UUID) {
+	var shipments models.MTOShipments
+	ctx.DB().Scope(utilities.ExcludeDeletedScope()).All(&shipments)
 }
 ```
 
+By passing in the model we can also derive the table name if needed to disambiguate from other tables.
```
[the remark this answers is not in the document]
duncan-truss: Yea that definitely needs a comma at the very least.  Part of my thinking is the query error you would get from not specify the model would be an "ambiguous column" error so I was trying to use the same language they might see coming back from the database.  Let me rework it.

docs/backend/guides/how-to/soft-delete.md
duncan-truss: Ok let me see if I can write up something clearer.  I think once you start mixing good examples and bad examples sometimes the reader will skim them blindly without realizing that's the thing they shouldn't be doing.

ronaktruss: To emphasize common errors or add examples of things to avoid it might be worth utilizing [admonitions](https://docusaurus.io/docs/markdown-features/admonitions). This might be relevant to the other comment as well.

duncan-truss: Cool, I haven't come across admonitions before!  Looks like I can nest a code block in there too but will try not to use them EVERYWHERE.

docs/backend/guides/how-to/soft-delete.md
```diff
@@ -26,16 +26,59 @@ type ExampleModel struct {
...
+    var documents models.Documents
+    ctx.DB().Scope(utilities.ExcludeDeletedScope(models.Document{}, models.UserUpload{})).
+        Join("user_uploads", "user_uploads.document_id = documents.id").
+        All(&documents)
+}
+```
+
+:::caution
+
+Unfortunately this will not filter any eager loaded associations so you may need to continue doing that depending on your query.
```
ronaktruss, line 55: Nit: `continue doing that` is a bit vague. I assume this is referring to filtering, but we should be explicit here.

ronaktruss (approved):
LGTM. Thanks for adding this documentation! 🎊

docs/backend/guides/how-to/soft-delete.md
duncan-truss: Mmm yea that assumes an instance where you were already filtering after the fact as opposed to a new case.

PR #133, opened 2022-05-12
Update creating an endpoint documentation to reflect newer pattern

YanZ777:
The documentation was referencing `AuditableAppContextFromRequest` instead of `AuditableAppContextFromRequestWithErrors`.


michaelvaldes-truss (approved):
thanks for catching this!

rogeruiz:
Yeah this documentation is very very old. There's also a documentation update that needs to happen once `AppContextFromRequest` is removed or made private.

PR #137, opened 2022-06-08
MB-12451 Update Service Object Docs to Include Orchestrator Info and Restructure Service Object Docs

felipe-lee:
[JIRA ticket](https://dp3.atlassian.net/browse/MB-12451)

The main purpose of this PR was to add information on orchestrator service objects, but while doing that, I noticed a few issues with the docs and so I updated quite a bit more...details in the change log below.

### Change Log

- [x] Create folder to contain all service object docs
- [x] Break down existing docs (2 very large pages) into smaller, more digestible pages. This also serves to make it easier to go to just the part you care about if you only need one part of it.
    - In part I'm also re-structuring the docs to work more like a start to end tutorial of sorts.
- [x] Switch from using `Reweigh` service objects as examples to using a fake `Pet` model to create the service objects. 
    - The main drivers for this was to make it easier to understand without needing to understand the underlying model too deeply, and because the `Reweigh` service objects no longer follow this pattern in the repo, which can be confusing for people looking at the docs and repo together. 
- [x] Document different existing patterns and suggest one for now. Will need to get feedback on this from more people though.
- [x] Add examples of tests for all parts of the service objects and related code.

### Viewing Docs

There's a lot here so the easiest way to view this is probably to pull down the repo and [run the docs locally](https://github.com/transcom/mymove-docs#running-locally-on-macos).

Then you can visit http://localhost:4000/mymove-docs/docs/backend/guides/service-objects/overview which is the entry point to the new docs


docs/backend/guides/service-objects/getting-started.md
```diff
@@ -0,0 +1,51 @@
...
+These docs will go through the process of creating a new set of service objects, but the info in each section should 
+be applicable whether you're creating a new service object or editing an existing one. We'll be focusing on creating 
+a data-type service object, but the information should be applicable to other types of service objects. We'll also 
+be using a fictional model and set of service objects to make it less likely that there will be drift between what 
+you see here and what is in the `mymove` codebase (since none of the code here will actually be in the codebase).
+
+## Fictional Model
+
+For reference, here is the fictional model we'll be creating service objects for (from our 
+[docs on db migrations](/docs/backend/setup/database-migrations)):
```
pearl-truss, line 15: Praise: I love this tie in to the migrations example data

docs/backend/guides/service-objects/structure.md
```diff
@@ -0,0 +1,145 @@
...
+re-use across service objects.
+
+## Orchestrator Service Objects High Level Structure
+
+Orchestrator service objects are similar overall to the other types of service objects, but the top level structure 
+and naming are slightly different. 
+
+We have an `orchestrators` subpackage inside `services` that holds the subpackages for each orchestrator. This helps 
+us separate out the `orchestrators` from the other more "regular" service objects (i.e. pre-existing service object 
+types that existed before `orchestrators` were added).
```
pearl-truss, line 74: **Question**: I'm reading this as the difference between an orchestrator and a regular service object is whether it's a new service object or not? Asking because I thought there were other differences between the two?

docs/backend/guides/service-objects/structure.md
```diff
@@ -0,0 +1,145 @@
...
+│   │   ├── order/
+│   │   ├── ...
+│   │   ├── reweight.go
+│   │   ├── shipment_orchestrator.go  <- the interface in the services package for a single orchestrator
+│   │   ├── sit_extension.go
+│   │   ├── ...
+```
+
+So in our example above, we have an orchestrator service object called `shipment` that lives in the `orchestrators` 
+subpackage. It has a corresponding interface file called `shipment_orchestrator` in the `services` package.
```
pearl-truss, line 96: **Suggestion**: It could be helpful to relate the orchestrator concept back to the Cat example from earlier

docs/backend/guides/service-objects/structure.md
felipe-lee: Orchestrators are still their own service objects, but the main differences are their placement (inside `orchestrators` subpackage), naming (orchestrators have a bit less stuttering, as our docs call it, and the interface file ends with `_orchestrator.go`), and what type of work they do (light audits and mostly chaining other service objects appropriately). 

Thanks for bringing it up though, I think I can maybe highlight some of these differences more in a few parts.

docs/backend/guides/service-objects/structure.md
felipe-lee: I think this would be great! Do you happen to have an example in mind? The main thing I can think of right now would be if we tweaked the sample `Cat` model to be a child model of a more generic `Pet` model. That could maybe also set us up to have good examples of setting up an orchestrator as well.

felipe-lee: I ended up moving toward a `Pet`/`Cat` parent/child models and updated the docs. This should hopefully make it easier to show orchestrators and to note their differences.

docs/tools/mockery/generate-mocks-with-mockery.md
```diff
@@ -20,3 +20,42 @@ type AccessCodeClaimer interface {
     ClaimAccessCode(code string, serviceMemberID uuid.UUID) (*models.AccessCode, *validate.Errors, error)
 }
 ```
+
+## Using Mocks
```
felipe-lee, line 24: This section is pretty much the same as before, I just moved it from the service object docs to this page since it fits better here.

docs/backend/setup/database-migrations.md
```diff
@@ -101,7 +101,7 @@ When you are writing the code for your new migration, there are a few things you
 This section covers what it's like to add a new table, though some information is helpful when editing existing 
 tables/models as well.
 
-For the examples in this section, we'll be adding a new model called `Cat`, with a corresponding table called `cats`.
+For the examples in this section, we'll be adding a new model called `Pet`, with a corresponding table called `pets`.
```
felipe-lee, line 104: The changes to this file are primarily to support the changes Pearl and I talked about in https://github.com/transcom/mymove-docs/pull/137#discussion_r896084859

These aren't quite necessary, but they do make it so that these models are the same across the docs and add a few examples of things to this page that weren't available with the previous version of the model

pearl-truss (approved):

docs/backend/guides/service-objects/set-up-service-subpackage-and-interface.md
```diff
@@ -0,0 +1,214 @@
...
+Following our naming conventions stated in
+[Structure - Service Object Naming](/docs/backend/guides/service-objects/structure#service-object-naming), we'll 
+define the interface functions like so:
+
+```go title="pkg/services/pet.go"
+package services
+
+// PetCreator Interface for the service object that creates a pet
+type PetCreator interface {
+	CreatePet() (
```
ronaktruss, line 60: missing closing paren
```suggestion
	CreatePet() ()
```

docs/backend/guides/service-objects/implementation.md
```diff
@@ -0,0 +1,1099 @@
...
+}
+
+// NewOfficePetUpdater creates a new petUpdater struct with the checks it needs for an office user
+func NewOfficePetUpdater() services.PetUpdater {
+	return &petUpdater{
+		checks: officeChecks(),
+	}
+}
+
+// UpdatePet creates a pet
```
ronaktruss, line 522: Comment might have not been updated from the creator
```suggestion
// UpdatePet updates an existing pet
```

docs/backend/guides/service-objects/overview.md
```diff
@@ -0,0 +1,79 @@
...
+1. **Presentation layer.** - Our endpoint handlers that perform type conversions and connect user requests to 
+   business logic functions.
+2. **Business logic layer.** - Code that implements MilMove's business logic.
+3. **Data access layer.** - Code that directly interacts with and manipulates the database.
+
+The `services` package is a combination of the bottom two layers, **business logic** and **data access**. It is 
+located within the [`mymove` `./pkg/services`](https://github.com/transcom/mymove/tree/master/pkg/services) directory.
+
+Our "service objects," as we call them (this is MilMove-specific terminology), are the structs/functions within this 
+package that implement our business logic. We have a few different kinds of service objects:
```
shimonacarvalho, line 20: Since the person reading "What is a service object" might literally have no clue what one is, I think it would be good to say "like AddressUpdater, MoveCreator". Otherwise this is very abstract and hard to connect to specific code.

docs/backend/guides/service-objects/overview.md
```diff
@@ -0,0 +1,79 @@
...
+
+- data type: These are the service objects that are tied to a specific model, e.g. logic related to the `Reweigh` 
+  model is in the `pkg/services/reweigh` service.
```
shimonacarvalho, line 23: Here we could again name specific service objects, rather than naming the models and packages they are associated with or located in. 

* data type - These are the service objects that are tied to a specific model, like `MTOShipmentUpdater`. It acts on the `shipment` model and located in the `pkgs/services/MTOShipment` service. 

Similarly for the other examples.

docs/backend/guides/service-objects/overview.md
```diff
@@ -0,0 +1,79 @@
...
+- orchestrators: These are service objects that help manage other service objects that are closely related, e.g. the 
+  logic that manages `MTOShipment` and `PPMShipment` service objects lives in `pkg/services/orchestrators/shipment`
```
shimonacarvalho, line 25: This one too - a little more clarity would be good because `MTOShipment` and `PPMShipment` are not service objects, but this description makes it sound like they might be.

docs/backend/guides/service-objects/validation.md
```diff
@@ -0,0 +1,1148 @@
...
+```
+
+</details>
+
+## Recap
+
+So now we've written rules for what validation we expect to happen with our data, and we've set up the validation 
+function that we can call to run through those rules. Now you can move on to setting up the actual service object.
+
+## Examples
```
YanZ777, line 1118: **question (non blocking):** Are there supposed to be two `##Examples` headings here?

YanZ777 (approved):
Thank you for writing all this documentation up! It's quite a lot. 
I like for the validators section you went TDD with the examples by showing the tests first and then the code. 
As someone who wasn't super familiar with Service Objects, these helped me understand them better and will be a great resource for me in the future. 
Looks good to me!

docs/backend/guides/service-objects/set-up-service-subpackage-and-interface.md
felipe-lee: Ah, nice catch, thanks, I had copy/pasted a more complete version and then trimmed it down to be barebones, but I guess I trimmed too much ✂️

docs/backend/guides/service-objects/validation.md
felipe-lee: Nope, thanks for calling it out!

docs/backend/guides/service-objects/overview.md
felipe-lee: Good point, I'll add this!

docs/backend/guides/service-objects/overview.md
felipe-lee: Good idea, thanks!

docs/backend/guides/service-objects/overview.md
```diff
@@ -20,6 +20,22 @@ Our "service objects," as we call them (this is MilMove-specific terminology), a
 package that implement our business logic. An example of a service object would be something like `AddressUpdater` 
 to update an `Address` record, or `MTOShipmentCreator` to create an `MTOShipment` record. 
 
+### Where Do Service Objects Fit In?
```
felipe-lee, line 23: @ronaktruss, we'd talked about this in one of our calls. Does this section get at what you were asking for?

docs/backend/guides/service-objects/overview.md
felipe-lee: I've updated the docs to have examples like this. Lmk if it looks better, or if you think more is needed. Relevant commit: https://github.com/transcom/mymove-docs/pull/137/commits/b0e1d7362b02f3ed778cee1272e93140ecfbd5ee

docs/backend/guides/service-objects/overview.md
felipe-lee: I've updated the docs to have examples like this. Lmk if it looks better, or if you think more is needed. Relevant commit: https://github.com/transcom/mymove-docs/pull/137/commits/b0e1d7362b02f3ed778cee1272e93140ecfbd5ee

docs/backend/guides/service-objects/overview.md
felipe-lee: I tried updating this section with more info on what the actual service objects are and the purpose of the orchestrator. Lmk if it looks better, or if you think more is needed. Relevant commit: https://github.com/transcom/mymove-docs/pull/137/commits/b0e1d7362b02f3ed778cee1272e93140ecfbd5ee

PR #139, opened 2022-06-16
Add a dependabot.yml to keep node and gh action deps up to date

felipe-lee:
Noticed we don't have anything to check on our dependencies in this repo


gidjin (approved):
LGTM! Thanks for adding this!

PR #159, opened 2022-06-24
Organize the sidebar

rogeruiz:
This patch updates the sidebar and adds some refinements that I missed when
working on #155. Most of the work is on the `/docs/adrs` route in the sidebar.
There is also an update in the footer to better call out ADRs as well.


rogeruiz:
I'm leaving this as **Draft** actually since I think we can do a bit more UX here for writing ADRs. A concept I was thinking was including a page that breaks down steps for writing an ADR such as copying the template, what to name it, what Front Matter to include in it, etc. I'll get to this sometime next week and then set it **Ready for review** afterwards.

YanZ777 (approved):
Looks good to me!

PR #162, opened 2022-06-28
Make light theme oceanicNext

rogeruiz:
I chose to make the light theme a different dark theme because of the
colors we chose for our site and how highlighting lines in code blocks
seems to not work properly due to how the light themes render the dark
selection background.

I specifically chose oceanicNext due to the fact that it fits well with the
MilMove color scheme.


michaelvaldes-truss (approved):
I dig oceanic next 💯

PR #181, opened 2022-07-18
MB-13121 Remove unnecessary direct dependencies

felipe-lee:
We don't use them directly. Seems like they came from when we first set up `docusaurus`, but `docusaurus` lists the right versions it needs as a dependencies, so we don't need to keep up our own version tracking for it.

Replaces #150 and #154


pearl-truss (approved):
makes sense to me 🚀

PR #198, opened 2022-08-26
[ MB-13351 ] Update the Redocusaurus internal documentation

rogeruiz:
This patch series includes some minor house cleaning and some major changes to
how we use the plugin due to upgrades in the Redocusaurus plugin's
configuration. I discovered the need to update this stuff while reviewing work
for MB-13351 transcom/mymove#9118.

- [x] Add in some better headings to the admonitions
- [x] Update the help channel from `wg-` to `c-`
- [x] A little blurb how this affects viewing changes
- [x] Update viewing local changes example
  - The main reason why this patch series exists.
- [x] Warn folks that the diff output doesn't apply

### DIFF
--- a/docs/tools/docusaurus/redocusaurus.md
+++ b/docs/tools/docusaurus/redocusaurus.md
@@ -99,6 +99,14 @@ index 697ffef..30fdcc4 100644
          ],
 ```
 
+:::warning About the diff output above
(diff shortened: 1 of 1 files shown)


ssilverm (approved):
LGTM

PR #200, opened 2022-08-31
[ MB-11967 ] Update documentation for running Go tests

rogeruiz:
This has come up during the MB-11967 ticket. We're making the Go test
documentation better together.

Pairing with @mmayo-truss on this. Currently this is a work in progress with
more updates to this description coming in the future.


kctruss (approved):
👍

PR #206, opened 2022-09-06
Mention demo and loadtest environments in secure migrations section

duncan-truss:
When the certificate in the demo environment needed a secure migration created recently, we realized the docs weren't as complete as they could be.  They only mentioned the exp, stg, and prd environments and not demo and loadtest, so that's what I've added here.


felipe-lee (approved):
LGTM. I pointed to one other place that has the envs listed, but it also isn't strictly necessary there imo.

PR #208, opened 2022-09-13
MB-12577 SQL templating ADR

kctruss:
## Summary

Research and compare options for SQL template architecture in MilMove to improve the developer experience of writing SQL queries and to open opportunities for SQL linting/checking.

### DIFF
--- a/docs/adrs/0073-using-templates-for-sql-queries.md
+++ b/docs/adrs/0073-using-templates-for-sql-queries.md
@@ -1,5 +1,5 @@
 ---
-title: 00XX Use Templates for SQL Queries
+title: 0073 Use Templates for SQL Queries
 description: |
   Decision outcome: Establish a pattern for SQL query template artifacts
 ---
--- a/docs/adrs/0073-using-templates-for-sql-queries.md
+++ b/docs/adrs/0073-using-templates-for-sql-queries.md
@@ -26,14 +26,18 @@ In order to achieve the best return value with the least effort, the chosen appr
 * *Do nothing*
 * *Utilize a SQL Template Engine*
 * *Build out our own SQL Templating architecture*
-* **A hybrid approach using an external library with some tweaks of our own**
+* **A hybrid approach using an external library implemented to work with Pop**
 
 ## Decision Outcome
 
 <!-- * Chosen Alternative:  -->
-* Chosen Alternative: *A hybrid approach using an external library with some tweaks of our own*
-* Positive Outcomes: We will have a SQL templating pattern that is easy to implement without requiring major refactor, provides us with linting opportunities and better code visualization, and may be very useful in the future if we end up implementing additional complex queries.
-* Consequences: If this new templating is not utilized elsewhere we will likely continue defaulting to writing SQL queries as strings in the Go code, which is our current pattern.
+* Chosen Alternative: *A hybrid approach using an external library implemented to work with Pop*
+* Positive Outcomes: 
+  * We will have a SQL templating pattern that is easy to implement without requiring major refactor, provides us with linting opportunities and better code visualization, and may be very useful in the future if we end up implementing additional complex queries.
+* Consequences: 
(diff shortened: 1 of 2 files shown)


rogeruiz:
This is amazing work @kctruss. I'll be review this more thoroughly for the next 20m but what I've read so far is great!

docs/adrs/0073-using-templates-for-sql-queries.md
```diff
@@ -0,0 +1,79 @@
...
+* **A hybrid approach using an external library implemented to work with Pop**
+
+## Decision Outcome
+
+<!-- * Chosen Alternative:  -->
+* Chosen Alternative: *A hybrid approach using an external library implemented to work with Pop*
+* Positive Outcomes: 
+  * We will have a SQL templating pattern that is easy to implement without requiring major refactor, provides us with linting opportunities and better code visualization, and may be very useful in the future if we end up implementing additional complex queries.
+* Consequences: 
+  * If this new templating is not utilized elsewhere we will likely continue defaulting to writing SQL queries as strings in the Go code, which is our current pattern.
```
rogeruiz, line 38: Love this breakdown right here. This is why I added @transcom/truss-db as an additional reviewer so others can see this even if it is after getting it merged in.

docs/adrs/0073-using-templates-for-sql-queries.md
```diff
@@ -0,0 +1,79 @@
...
+  * `+` Once established, future SQL query work following the determined pattern should
+  be more consistent, and have positive impacts for onboarding efforts.
+
+### *A hybrid approach using an external library implemented to work with Pop*
+
+* `+` *Simple setup.* 
+* `+` *The level of effort necessary to initiate refactor is low.*
+* `+` *The syntax is simple and easy to learn.*
+  * `+` Ongoing SQL work should be more consistent, and have positive impacts
+  for onboarding efforts.
```
rogeruiz, line 75: Love to see this! 🎉

rogeruiz:
Thought about adding _ready to merge_ since the number of the ADR could be affected by other ADRs currently in flight. I think we should get this merged in quickly. But I will let you decide that at your discretion, @kctruss

docs/adrs/0073-using-templates-for-sql-queries.md
```diff
@@ -0,0 +1,79 @@
...
+## Decision Outcome
+
+<!-- * Chosen Alternative:  -->
+* Chosen Alternative: *A hybrid approach using an external library implemented to work with Pop*
+* Positive Outcomes: 
+  * We will have a SQL templating pattern that is easy to implement without requiring major refactor, provides us with linting opportunities and better code visualization, and may be very useful in the future if we end up implementing additional complex queries.
+* Consequences: 
+  * If this new templating is not utilized elsewhere we will likely continue defaulting to writing SQL queries as strings in the Go code, which is our current pattern. 
+* Other considerations:
+  * The templating pattern must be clearly documented in order to be easily adopted.
```
rogeruiz, line 40: This doesn't need to be addressed here, but how would you like to approach this? We definitely need to socialize this more. I suggest [the #g-database channel in DP3 Slack](https://ustcdp3.slack.com/archives/CSGDM3NUW)🔒.

rogeruiz (approved):
lgtm 🌈 

Great work on this. I highlighted some additional points below.

docs/adrs/0073-using-templates-for-sql-queries.md
kctruss: I was thinking I would follow up with another PR adding a "how to" documentation article to docusaurus, do you think that makes the most sense?

rogeruiz: absolutely. But let's hold off on this since you've made the post in Slack about socializing it more. We can certainly tackle it in another sprint down the road. I think your ADR is documentation enough. But I also have transcom/mymove#9207 that tries to add the minimal amount of documentation around this. I think we're good enough for now.

PR #212, opened 2022-09-20
Update ADR 0073 to record implementation findings

kctruss:
Updated the existing ADR for SQL templating to account for discoveries made during initial implementation and to better capture the final direction taken.


ruizajtruss (approved):
This looks great. Appreciate we're crossing out the old notes just to keep that history there which provides some good context; nice job!

PR #216, opened 2022-09-28
Add documentation for suite.PreloadData

mlm55:
Add documentation for `PreloadData` and remove references to `RunWithPreloadedData`

### DIFF
--- a/docs/backend/testing/running-server-tests-inside-a-transaction.md
+++ b/docs/backend/testing/running-server-tests-inside-a-transaction.md
@@ -175,18 +175,14 @@ If you use `Suite.Run`:
 
 Sometimes, it's too much of refactor to pull out all the setup, or the setup takes a very long time.
 
-This is not a great situation, but if you still want to keep some of the transactional performance benefits, you can use an alternate `Run` function called `RunWithPreloadedData`. This is a **less ideal pattern** for the following reasons:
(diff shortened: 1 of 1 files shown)


docs/backend/testing/running-server-tests-inside-a-transaction.md
```diff
@@ -175,18 +175,14 @@ If you use `Suite.Run`:
...
-* It rolls back each subtest, but doesn't use a new connection. So any changes made in the main test will be available in the subtests.
-
-It can be preferred in one case:
+`PreloadData` can be preferred in one case:
 
 * Where there is a large amount of necessary setup, like populating the rate engine for payment calculations. In this case it can be valuable to have a preloaded setup.
 
-So **if you absolutely must** use `RunWithPreloadedData`, the process is easy. Just change the `suite.Run` calls to `suite.RunWithPreloadedData`.
+So **if you absolutely must** use `PreloadData`, the process is easy. Create a `suite.PreloadData` function and do all your test setup inside. [Here's an example](https://github.com/transcom/mymove/blob/c4f7df208f4b63fd85b0cd7645156a3516902027/pkg/rateengine/nonlinehaul_test.go#L188)
+
```
shimonacarvalho, line 186: Looks good! 

Can you add one more bullet to the "My tests started failing after making these changes"?

Basically we do tell them to watch out for `suite.DB()`, but I'd like to make it clear that any call to `suite.DB()`, **even if they are creating a stubbed object**, should be moved from the parent test. Because it's the call to `suite.DB()` that performs some setup that we should only do in certain locations.

shimonacarvalho (changes requested):

shimonacarvalho (approved):

PR #224, opened 2022-10-20
Migrated manage dependabot to Confluence

YanZ777:
Remove the two articles about dependabot as well as any references to them. Delete the parent folder as well since it would be an empty folder.


rogeruiz (approved):
lgtm 🌈

thanks for this @YanZ777 verified that this documentation now exists in Confluence.

[=> In Confluence 🔒 ](https://dp3.atlassian.net/wiki/spaces/MT/pages/1921056804/Dependabot)

PR #231, opened 2022-11-01
MB-14102 - Add partial updates ADR

michaelvaldes-truss:
ADR 0076 explains the background behind needing to support partial updates and proposes a pattern for addressing this requirement:

There are two primary options that are being weighed in this ADR, an Intermediate Structs option, and a Handler Logic option, each of which has a proof-of-concept PR related to it.

Intermediate Structs:
[POC changes for Intermediate Structs proposal](https://github.com/transcom/mymove/pull/9503)
> In this proposal, a new service model, `MTOShipmentUpdate`, would serve communication between the handlers and the `UpdateShipment` service. 
The MTOShipmentUpdate model would represent an update to an MTOShipment. It would resemble the MTOShipment struct, but its fields could be Nullable, which means that they would contain information about whether a field should be updated by the update service.

Handler Logic:
[POC changes for Handler Logic proposal](https://github.com/transcom/mymove/pull/9858)
> In this proposal, the handler is responsible for fetching the db model to update, updating that struct in memory with information about which fields are present in the payload, and passing that updated struct to the service to perform validations and db operations.


docs/adrs/0076-partial-update-patches.md
```diff
@@ -0,0 +1,146 @@
...
+
+Such a design would have many advantages in addition to solving the problem of partial model updates and setting fields to null. The UpdateMTOShipment struct provides clarity by broadcasting intent: it clearly exists to update a shipment and it tells the service exactly what it wants. It also helps to further disentangle the handler layer from the database and services; since the handler now uses the MTOShipmentUpdate model, it no longer needs to know about the database structure. This design will also allow for a thorough cleanup of a lot of the handler and service logic through removal of repeated null checking and shifting business logic responsibilities cleanly to the services.
+
+### Implementation
+
+Due to the scope of work required, implementating this pattern would need to be done slowly.
+
+#### Decisions to finalize
+* Decide on a naming convention for intermediary update models
+* Determine if all models should fall under this pattern or only update models.
```
rogeruiz, line 104: I like these section, but I think in asking the questions you may have already answered them. The suggestion below is my re-wording, but there's no need to accept the changes.

```suggestion
#### Decisions to socialize
* The naming convention for intermediary update models will be `IntermediaryUpdateModels`.
* We'll do the `update` models first, but we will need measure if other models should also be a part of this similar pattern.
```

The reason I re-worded this is because I believe that even with a decision being made here, we can revisit any changes that need to be made in a subsequent ADR.

docs/adrs/0076-partial-update-patches.md
```diff
@@ -0,0 +1,146 @@
...
+> **Nullable field:** A field that contains information about whether it was included in the payload. A field with `Present=true` was included in the payload, with a `Value` that may or may not be null. A field with `Present=false` was not included in the payload, and therefore should not be updated in its corresponding model.
+
+```
+type NullableString struct {
+	Present bool // Present is true if key is present in json
+	Value   *string
+}
+```
+
+MTOShipments are an example of a case in which the back-end is not prepared to handle partial updates or intentionally setting fields to null. At the handler (e.g. the ghcapi [UpdateMTOShipmentHandler](https://github.com/transcom/mymove/blob/de9148371427ba348e94439b4acb766f05013797/pkg/handlers/ghcapi/mto_shipment.go#L284)), the payload (which may contain nullable types) is converted to an [MTOShipment model](https://github.com/transcom/mymove/blob/de9148371427ba348e94439b4acb766f05013797/pkg/handlers/ghcapi/internal/payloads/payload_to_model.go#L309), and during this conversion any information that distinguishes between missing fields and null fields is lost.
```
rogeruiz, line 23: Love this callout in the last sentence. Missing and null fields are different and should be treated differently. This is a great example of revising previous ADRs like _0066 Use custom nullable types for patch requests_.

rogeruiz (commented):
lgtm 🌈

Keep up the good work on this @kctruss and @michaelvaldes-truss! Y'all are moving in the right direction.

docs/adrs/0076-partial-update-patches.md
```diff
@@ -0,0 +1,146 @@
...
+
+#### Decisions to finalize
+* Decide on a naming convention for intermediary update models
+* Determine if all models should fall under this pattern or only update models.
+
+#### Refactor stages
+* Identify and refactor any patch endpoints that require nullable fields for deleting values.
+* Fully refactor related validator methods to cleanly handle intermediary update models
+* Refactor all update models to remove dependencies on the `models` package
+* Relocate fully refactored update models into a separate package for better separation of code
```
YanZ777, line 110: **questions:** Do these stages need to be done in order? Is there overlap between the different stages?

docs/adrs/0076-partial-update-patches.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+## Considered Alternatives
+* *Add a new changed values criterion for distinguishing between events when matching to event templates*
+* *Refactor our details types and details components so that different data can be displayed uniquely within the same details type*
+* *Do nothing*
+
+## Decision Outcome
+
+### Chosen Alternative: *Introduce a new struct that serves between the handler and service layers and broadcasts intent*
```
YanZ777, line 120: **request:** I'd like to hear more about what the pros and cons are as it relates to other parts of the application that aren't move history.

YanZ777 (commented):
I have no problems with the ADR in of itself, but I think getting a review from the folks I've tagged would be good.

docs/adrs/0076-partial-update-patches.md
michaelvaldes-truss: Oops, this was a copy/paste error. Fixed!

docs/adrs/0076-partial-update-patches.md
```diff
@@ -0,0 +1,146 @@
...
+* `-` *Composite struct models have dependencies on current `models` package which means they cannot be separated into a new package without introducing circular dependencies*
+
+## Pros and Cons of the Alternatives
+
+### *Update model to store information about "presence"*
+* `+` *Addresses issue of not being able to delete fields with patches*
+* `+` *Little added effort when updating model with additional fields*
+* `-` *Violates expectations of pop database models*
+* `-` *Large amount of work involved in refactoring current models and all services that use them*
+
```
shimonacarvalho, line 137: So there was no good way to make a model that would work for pop and support presence information?

Or there is but the lift is too great?

michaelvaldes-truss: In a way it's kind of both, but I think the "no good way to do it" argument is most compelling.
There's no good way to make a model to support presence information because presence information is beyond the scope of a database model. From what I understand, a DB model should just give information about the current state of a row in a DB at a certain point in time. What we're trying to capture here is more related to communicating information about presence from the handler to the service, which is the intention of the Update models.

docs/adrs/0076-partial-update-patches.md
kctruss: I completely agree @shimonacarvalho - I'd like to find a good naming convention that describes what these are and their intended use case. In researching design pattern opportunities, I also locally kicked around "Payload Models" and "API Models". Both assume that our current Models package implies "Data Models", however which can still feel confusing and/or misleading. I can see a use case for non-update payloads (i.e. I've worked in such a pattern in the past) but the scope of implementing a pattern like that would be enormous, which is why I've gone back and forth.

michaelvaldes-truss: These are great callouts @rogeruiz and @shimonacarvalho , thank you!
With respect to naming, the update structs are _structs_, but they are not specifically _database models_. I've changed the language of the ADR to reflect that.
As for the decisions to finalize, I've taken yall's advice and removed that section and answered those questions directly in the ADR.
The "Requirements and Naming" section should address all of the above.

docs/adrs/0076-partial-update-patches.md
michaelvaldes-truss: These stages mostly have to be done all at once to avoid build errors.
This section doesn't necessarily describe a required output of this ADR; I think it's included to describe the hypothetical work needed to refactor an existing model/endpoint/service system to work with an Update struct, based on our experience introducing the MTOShipmentUpdate in the POC branch linked in the description.

docs/adrs/0076-partial-update-patches.md
```diff
@@ -0,0 +1,149 @@
...
+	TACType                          *LOAType
+	SACType                          *LOAType
+	PPMShipment                      *PPMShipment
+	CreatedAt                        time.Time
+	UpdatedAt                        time.Time
+	DeletedAt                        *time.Time
+}
+```
+
+Handlers would be in charge of converting payloads into the MTOShipmentUpdate struct and passing it to the UpdateShipment service. The UpdateShipment service would be responsible for fetching the database model (MTOShipment) and updating it appropriately based on the `Present` fields in the MTOShipmentUpdate.
```
duncan-truss, line 94: I guess this is roughly what payload to model functions are doing currently but I don't think we want additional logic added into our handlers.  I think this might add another layer of confusion because not all fields are updatable by each API (customer, office, prime, ...).  Are you saying each handler will provide an API specific **intermediate** structs that calls a dedicated service method that restricts unintended fields being set/updated?

docs/adrs/0076-partial-update-patches.md
```diff
@@ -0,0 +1,149 @@
+---
+title: "0076 Partial Update Patches"
+---
+
+# *0076 Partial Update Patches*
+
+**🔒 User Story:** *[MB-14102](https://dp3.atlassian.net/browse/MB-14102)*
+
+At present, the MilMove back-end does not have developed support for partial updates or deleting fields in its models through patches. [ADR 0066](docs/adrs/0066-use-custom-nullable-types-for-patch-requests.md) introduced custom nullable types for patch requests which addresses this issue at the swagger/handler layer, but services may not be prepared to fully support the rest of this sort of workflow.
```
duncan-truss, line 9: I think the PPM shipment updater allows for partial updates. I think for PPMs we have some dependent fields that the service keeps track of clearing on updates like if an advance is changed and we need to clear the amount.

Is the other issue here that MTOShipment just has too many nullable fields because it is used to support HHG, NTS, and NTSR shipments and that really should be refactored?  How many fields really need to be nullable/deletable could we see a breakdown of how hard that would be to handle on a case by case basis in a merge function or dedicated endpoints? Is it partially that our frontends are sending nulls when maybe we should be clearer about treating omitted keys differently?

docs/adrs/0076-partial-update-patches.md
```diff
@@ -0,0 +1,149 @@
...
+* Refactor all update structs to remove dependencies on the `models` package
+* Relocate fully refactored update structs into a separate package for better separation of code
+
+
+## Considered Alternatives
+* *Introduce a new struct that serves between the handler and service layers and broadcasts intent*
+* *Update model to store information about "presence"*
+* *Pass payload to service*
+* *Do nothing*
+
```
felipe-lee, line 121: **thought**: I feel like another option (even if we say no to it), is having a flow kind of like this for update handlers:

1. Auth check as needed
1. Use fetcher service object to retrieve current model
1. Error check
1. Pass current model and payload to payload_to_model function
    *  This will update the current model with the data from the payload, so we could take into account when fields are present vs not, and if they are, when their value is `null`
    * The payload_to_model function wouldn't have any explicit checking of whether a thing _should_ be done or not, just what is possible to set on the model based on the payload info. Validation should happen in the service object.
1. Pass the current and "new" models to the updater service object.
    * Updater service object validates data, makes any other changes needed (e.g. calculating incentives as needed for PPM shipments), and makes the call to the db to save changes.
1. Error check
1. Pass newly updated model to model_to_payload func
1. Return response

I've outlined some rough pros/cons below, though 

Some pros:
* No need to introduce new structs
* This would standardize us a bit more on making sure to have fetcher service objects
* Wouldn't need to refactor service objects to understand a different struct.
* Would simplify updater service objects
    * Wouldn't need merge functions to try to figure out what the state should be (all those null checks mentioned in this ADR)
    * Depending on how we wanted to do fetching, this workflow might remove that logic entirely from the updater service object (see comments below in cons) 

Some cons:
* It would take some work to add fetcher service objects for things missing them
    * While the extra work now is a con, I think long-term, it's a pro that we would have these (hence the pro above)
* Handlers would need to have some more logic than they do now (at least the ones that essentially only call one service object to do most of the work). 
    * Is this blurring the line for separation of concerns? This could be seen as maybe the handler having too much of the work, but for this workflow, I see fetching as part of prepping to convert the payload to a model. In fact, we could even possibly move the fetch there, though that would require more refactoring, and would make it harder to inject dependencies in tests. It would also likely mean we'd want to be testing payload_to_model funcs more thoroughly than we do now, or rather, test them to begin with 😅 
* Would require us to refactor service objects to accept both the current and new versions of the model
    * Part of this refactor would also be removing the fetch.
    * Alternatively, we could even make the service objects not trust the current model that the handler fetched (nor accept it as an arg) and instead re-fetch the data.
        * I feel like from the perspective of ensuring that everything is correct, I would almost like to make the service objects re-fetch, but this would also mean extra queries, which might not be desirable.
        * A possible reason to do this would be to make the fetch call that the handler does only include fields that are directly on the model, while the fetch in the service object could fetch related models too since they may be needed for some things (e.g. calculating final incentive for a ppm shipment requires we load related weight ticket objects).

ahobson: From my perspective, the key insight of this ADR is that models are not a good representation of what should go over the wire for the API contract. It seems to me like the service is the actual contract we are trying to expose.

I personally think we should have something like `payload_to_service_request` that translates the swagger payload to an internal service request object. Then, as Felipe says above, the service can then control fetching the model with all the fields it needs and do whatever it needs to do. The service then can build a service response object which the handler then uses to call `service_response_to_payload`.

I think this is something we need because some API requests only accept certain fields of the model or certain combination of properties in the model. Same for the responses. As it exists today, our swagger definitions do not match reality. They also create confusion when creating an new object (e.g. service item). Is the client allowed to provide an ID? Our current swagger definitions allow for it. When updating, which fields are allowed to change?

If we had explicit service object request and response objects, then models would no longer be exposed to handlers (or the API)! We would be able to explicitly document via swagger and in the go code which fields are allowed to change.

The downside, of course, is that is not the pattern we have today and conversion would be time consuming and probably never actually finished.

I wonder what you all think of these ideas

felipe-lee: What would the "internal service request object" look like though? Would it be like the struct in the currently chosen solution for the ADR? Like it kind of sounds like they're similar objects in that they don't really match the api or model, they match what the service objects need in order to do their work.

I definitely agree that we should be better about making sure our API spec actually matches reality. We have so many places that don't actually list the required fields, or only have a partial list. Same with misleading fields like it seeming like ID is editable.

docs/adrs/0076-partial-update-patches.md
michaelvaldes-truss: In this proposal, the MTOShipmentUpdate struct pretty much takes the place of the MTOShipment model in these PayloadToModel functions that the handlers use, and they can all share this intermediate struct without increasing the complexity of each handler by much at all.

docs/adrs/0076-partial-update-patches.md
michaelvaldes-truss: I imagine the internal service request struct would vary based on the operation/endpoint. This ADR attempts to achieve this in a way that is a little less devastating to our current code by using a struct that greatly resembles the MTOShipment, and which can be easily updated to make additional fields nullable.
However, if we wanted to go in the direction of stricter payload_to_service_request objects, we would need a much bigger refactor than this.

docs/adrs/0076-partial-update-patches.md
michaelvaldes-truss: This is a great point, and I would say that this ADR attempts to introduce this pattern as an option for our endpoints in cases where it would be beneficial. I don't think it's necessary or even helpful to refactor all of our endpoints, but some fields in the MTOShipment that should be nullable aren't currently nullable, and future models might also have this problem.
I also think that regardless of what happens in the frontend, as long as we're using our database models in the request payloads (and convert to them in the handlers), this problem will exist without some sort of big refactor.

rogeruiz: >I don't think it's necessary or even helpful to refactor all of our endpoints, but some fields in the MTOShipment that should be nullable aren't currently nullable, and future models might also have this problem.

I'd suggest adding this to ADR to make this point clear. Also make sure to state earlier in the ADR that the MTOShipment fields are the narrow focus here. That could help make approval on this ADR easier.

docs/adrs/0076-partial-update-patches.md
rogeruiz: >they can all share this intermediate struct without increasing the complexity of each handler by much at all.

can you clarify this section in the ADR a bit more to explain that this wouldn't be adding additional logic added to our handlers?

docs/adrs/0076-partial-update-patches.md
felipe-lee: Ah, yeah that does seem like it would require more work. Though one upside to it is that it would make it easier for us to more fully decouple our models from the API spec since currently a lot of our payloads match the models closely, even when they probably shouldn't.

@michaelvaldes-truss, what do you think about adding the option I mentioned? I personally would kind of prefer that than having more structs, but even if we go with the currently proposed decision, it might be good to have this in the ADR so that future peeps know we did consider it and chose something else.

michaelvaldes-truss: Absolutely -- I'll review that option in more detail and add it to the ADR!

michaelvaldes-truss: @felipe-lee I added a section to the ADR for your proposal, and even started a POC branch to demonstrate how it would look. 
I really like this option (which I'll call the Handler Logic option from here) for its upsides, mainly that it's a lot easier to implement than the Intermediate Structs option and that it doesn't overcomplicate the service validators.
One main drawback, as you and @ahobson mentioned, is that it still keeps our handlers/endpoints tightly coupled to our database models and keeps/moves a lot of service-ish logic to the handlers.

rogeruiz:
👍🏽 vote for Intermediate Structs
- https://github.com/transcom/mymove/pull/9503

👎🏽 vote for Handler Logic
- https://github.com/transcom/mymove/pull/9858

rogeruiz:
I'm reviewing this today during my office hours. I resolved some comments that were old at this point to make it easier to review this. Please un-resolve comments if I mistakenly resolved an active comment. Thanks! Final review should be done by 12:00 ET today.

rogeruiz (approved):
lgtm 🌈

I've reviewed this ADR and the main parts that stand out to me are the following pros and cons.

For the **Handler transformation**:
> Doesn't overcomplicate service validators

Against the **Intermediate structs**
> Composite struct models have dependencies on current models package which means they cannot be separated into a new package without introducing circular dependencies

Since the **Handler transformation** option is being chosen here, this means we should be closing transcom/mymove#9503 and working towards getting transcom/mymove#9858 into a reviewable and mergable state.

docs/adrs/0076-partial-update-patches.md
```diff
@@ -115,24 +118,41 @@ Due to the scope of work required, implementating this pattern would need to be
...
 * *Update model to store information about "presence"*
 * *Pass payload to service*
 * *Do nothing*
 
 ## Decision Outcome
+### Chosen Option: **Handler Transformation** -- move any logic for converting the payload to a model from the service to the handler
+* `+` *Addresses issue of not being able to delete fields with patches*
+* `+` *Relatively simple to implement in existing handlers*
+* `+` *Doesn't overcomplicate service validators*
+* `-` *Handlers responsible for a lot of logic, including fetching db model and updating that fetched struct with payload intelligently*
```
felipe-lee, line 145: **thought**: I think that while this is true, it could be misread. Like, yes, this proposes we move the fetch to the handler, but not directly. Handlers should use fetcher services to get the current version rather than having an inline fetch. I don't know if we need to make that distinction here or not, but just wanted to call that out.

felipe-lee (approved):
I left a comment, but I don't feel very strongly about it. This lgtm, thanks for all the work y'all have put into this!

PR #233, opened 2022-11-03
Create an ADR for using YAC Spin

YanZ777:
This adds an ADR for using YAC Spin instead of pterm's spinner. 
Pterm will not be completely replaced as YAC Spin only deals with the spinner.

### DIFF
--- a/docs/adrs/0075-cli-spinner.md
+++ b/docs/adrs/0075-cli-spinner.md
@@ -1,0 +1,45 @@
+---
+title: "0075 Use YAC Spin"
+---
+
+# _Use YAC Spin_
+
+A CLI spinner is a spinner that appears on the command line interface to visually indicate that the program being run is working on a task.
+The MilMove project uses such spinners when loading or parsing documents. pterm spinner, the current spinner being used on MilMove, is not thread safe. This was discovered while working on [MB-12825 Race Conditions](ttps://dp3.atlassian.net/browse/MB-12825). pterm offers many other command line interface features that are thread safe; it is specifically the spinner that generates warnings when Go's `-race` flag is turned on while running tests.
+
+## Decision Drivers
+
+- We want to resolve the race conditions that appear when the `-race` flag is turned on in Go
+
+## Considered Decisions
+
+- Use [YAC Spin](https://github.com/theckman/yacspin)
+- Replace [pterm](https://github.com/pterm/pterm) completely
+- Continue to use [pterm](https://github.com/pterm/pterm) spinner
+
+## Decision Outcome
+
+- Chosen Decision: Use YAC Spin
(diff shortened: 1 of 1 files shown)


rogeruiz:
Adding myself as Assignee to review this next week. Thanks @YanZ777

reggieriser (commented):
Nice job on discovery of the issue and the ADR.  I left this earlier in a Slack thread, but will repeat it here:

I think I was the one that added `pterm` and the “pretty” output to the pricing parser/import stuff quite a while ago.  The spinner is just eye candy and should be able to be removed fairly easily (we will still want the text of the spinner to remain, though, since it provides good context).  Everything should be single-threaded when importing via the `ghc-pricing-parser` CLI so I suspect the race condition is just an artifact of tests running in parallel.  As a test, I tried removing the three spinners we have and then running all tests with `-race` and no race conditions were noted.  I’d support getting rid of the spinners entirely.

YanZ777:
> Nice job on discovery of the issue and the ADR. I left this earlier in a Slack thread, but will repeat it here:
> 
> I think I was the one that added `pterm` and the “pretty” output to the pricing parser/import stuff quite a while ago. The spinner is just eye candy and should be able to be removed fairly easily (we will still want the text of the spinner to remain, though, since it provides good context). Everything should be single-threaded when importing via the `ghc-pricing-parser` CLI so I suspect the race condition is just an artifact of tests running in parallel. As a test, I tried removing the three spinners we have and then running all tests with `-race` and no race conditions were noted. I’d support getting rid of the spinners entirely.

Thanks for reviewing this! I'm going with the decision to remove the spinner code. 
I think you're correct on the race conditions are an artifact of tests being run in parallel.

docs/adrs/0075-cli-spinner.md
```diff
@@ -0,0 +1,54 @@
...
+- Use [YAC Spin](https://github.com/theckman/yacspin)
+- Replace [pterm](https://github.com/pterm/pterm) completely
+- Continue to use [pterm](https://github.com/pterm/pterm) spinner
+
+## Decision Outcome
+
+Remove all spinner code.
+
+- This allows us to resolve races conditions that are flagged by Go's `-race` flag. Resolving these race conditions helps resolve [STIG V-70185](https://www.stigviewer.com/stig/application_security_and_development/2018-12-24/finding/V-70185).
+- Having a spinner display is not critical to the operation of the application. The text of the messeage should be retained to maintain context as to what is being run.
```
reggieriser, line 26: **nitpick:** `messeage` -> `message`

reggieriser (approved):
LGTM!  Thanks for adding the additional option of removing the spinner (and doing that work!).

PR #242, opened 2022-11-29
Switch search plugin

shimonacarvalho:
I found that there's another version of the docusearch local plugin - tried it out and lo and behold, it does all the things that seem to be broken on our own version. 

* Searches page titles by search string - I could never find the "Running server tests" page in the current docusaurus but now I can!
* Shows the page title on the search list - helps contextualize when you have a number of identical hits. 
* Provides a "see all results" link to see all the hits instead of just the top few.

Generally it seems to return a lot more results and give a lot more context for the searcher. 
Reviewer: please let me know if there's any other processes to remove one plugin/package with yarn and install another.

Check it out!
image

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -181,8 +181,19 @@ module.exports = {
       }
     ],
   ],
+  themes: [
+    [
+      require.resolve("@easyops-cn/docusaurus-search-local"),
+      /** @type {import("@easyops-cn/docusaurus-search-local").PluginOptions} */
+      ({
+        // `hashed` is recommended as long-term-cache of index file is possible.
+        hashed: true,
+        indexBlog: false,
+      }),
+    ],
+  ],
+
   plugins: [
-    require.resolve('@cmfcmf/docusaurus-search-local'),
     [
       '@docusaurus/plugin-client-redirects',
       {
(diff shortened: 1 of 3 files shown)


YanZ777 (commented):
The changes themselves look good, thanks for making them. 
I'm getting a different `yarn.lock` when I regenerate it though. 
I did
```
rm yarn.lock
yarn install
```
And that is giving me a different file than the one in this PR. How did you regenerate the `yarn.lock`?

shimonacarvalho:
> And that is giving me a different file than the one in this PR. How did you regenerate the `yarn.lock`?

I updated it - is this one better? Do you know why there's all that algolia stuff in there - we are not using algolia.

YanZ777:
> > And that is giving me a different file than the one in this PR. How did you regenerate the `yarn.lock`?
> 
> I updated it - is this one better? Do you know why there's all that algolia stuff in there - we are not using algolia.

Yep, this looks good! 
Running `npm list -a` (and doing an output redirect to a file since it's a lot of output), it looks like the algolia stuff is coming from `"@docusaurus/preset-classic"`. 
I'm guessing prior folks who added dependencies forgot to delete the `yarn.lock` and do `yarn install`.

YanZ777 (approved):

PR #245, opened 2022-12-01
Use frozen lockfile when installing dependencies

YanZ777:
We should be using a frozen lockfile when installing dependencies for our CircleCI jobs. 
Using a frozen lockfile will help with detecting if a dependency has been added or removed without updating the lockfile since the job will fail if the correct dependencies are not installed.


rogeruiz (approved):
lgtm 🌈

Good catch on this @YanZ777 thank you!

PR #252, opened 2022-12-16
Add article on understanding milmove routing

YanZ777:
Add an article that describes how the routing on MilMove is set up.

### DIFF
--- a/docs/backend/guides/understanding-milmove-routing.md
+++ b/docs/backend/guides/understanding-milmove-routing.md
@@ -10,7 +10,7 @@ MilMove is using [Gorilla Mux](https://github.com/gorilla/mux) for routing.
 ## Code Snippet Analysis
 
 :::note
-This document won't analyze all the code related to how server side routing works, but select ones that are considered notable and for which there is extensive knowlege on.
+This document won't analyze all the code related to how server side routing works, but select ones that are considered notable and for which there is extensive knowlege.
 :::
 
 ### Subrouting
--- a/docs/backend/guides/understanding-milmove-routing.md
+++ b/docs/backend/guides/understanding-milmove-routing.md
@@ -1,0 +1,115 @@
+# Understanding MilMove Routing
+
+This document is a brief summary on how routing on MilMove works.
+
+## Routing
+
+The routing is initialized in `serve.go`, which calls `InitRouting` inside of `routing_init.go`. This initializes the routing that will be used by the MilMove application. This sets up routing for all 3 sites, the MilMove application, office, and admin. Perhaps this not ideal, but this is how it works as of the writing of this document (December 2022).
+MilMove is using [Gorilla Mux](https://github.com/gorilla/mux) for routing.
+
+## Code Snippet Analysis
+
+:::note
+This document won't analyze all the code related to how server side routing works, but select ones that are considered notable and for which there is extensive knowlege on.
+:::
+
+### Subrouting
+
+Gorilla Mux allows easy configuration of matching routes.
+Here were are matching by a path prefix. This means that any requests that are prefixed with `/static`, will use the set middleware and handlers.
+
+Here's the code with some comments to further explain.
+
+```golang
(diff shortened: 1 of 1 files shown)


docs/backend/guides/understanding-milmove-routing.md
```diff
@@ -0,0 +1,115 @@
+# Understanding MilMove Routing
+
+This document is a brief summary on how routing on MilMove works.
+
+## Routing
+
+The routing is initialized in `serve.go`, which calls `InitRouting` inside of `routing_init.go`. This initializes the routing that will be used by the MilMove application. This sets up routing for all 3 sites, the MilMove application, office, and admin. Perhaps this not ideal, but this is how it works as of the writing of this document (December 2022).
```
kctruss, line 7: > Perhaps this is not ideal, but this is how it works

🤣 I appreciate the honesty

docs/backend/guides/understanding-milmove-routing.md
```diff
@@ -0,0 +1,115 @@
...
+MilMove is using [Gorilla Mux](https://github.com/gorilla/mux) for routing.
+
+## Code Snippet Analysis
+
+:::note
+This document won't analyze all the code related to how server side routing works, but select ones that are considered notable and for which there is extensive knowlege on.
```
kctruss, line 13: **nitpick:** the last "on" in the sentence is not needed. #GrammarNerd

kctruss (approved):
This is so good and hugely helpful, Yan, thank you! I appreciate the way you break down information in an accessible and easy-to-digest way.

docs/backend/guides/understanding-milmove-routing.md
YanZ777: I appreciate the grammar nerd!

docs/backend/guides/understanding-milmove-routing.md
shimonacarvalho, line 7: Might be helpful to add, it also sets up routing for Prime API and Support API?

docs/backend/guides/understanding-milmove-routing.md
```diff
@@ -0,0 +1,115 @@
...
+## Code Snippet Analysis
+
+:::note
+This document won't analyze all the code related to how server side routing works, but select ones that are considered notable and for which there is extensive knowlege.
+:::
+
+### Subrouting
+
+Gorilla Mux allows easy configuration of matching routes.
+Here were are matching by a path prefix. This means that any requests that are prefixed with `/static`, will use the set middleware and handlers.
```
shimonacarvalho, line 19: "Here we are"?

docs/backend/guides/understanding-milmove-routing.md
shimonacarvalho: Also, I like to sometimes add a simple definition/link to get people started if they really are new to a concept. 

"Routing is a mechanism where HTTP requests are routed to the code that handles them. It is how we determine what should happen when a user visits a certain URL."

docs/backend/guides/understanding-milmove-routing.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+All middleware follows the format of returning a function which takes in an `http.Handler` and returns an `http.Handler`.
+Inside this closure, one can do whatever one desires to do using the `http.ResponseWriter` and `http.Request`.
+In this case for `ValidMethodsStatic`, the function checks if the request is either a `GET` or a `HEAD` request, and returns an error if the request is neither. Afterwards, it calls the next handler.
+
+### Index Handler
+
+This index handler is used to handle all path prefixes that start with `/`.
+
```
shimonacarvalho, line 76: On a second reading I actually don't understand what this means. 

Would be nice to provide an example to make this clearer. Without looking it up, I'm not sure what would be a path prefix that starts with `/`. Is that like `https://my.stg.move.mil/` ?

YanZ777:
Putting up the follow up suggestions into a new PR, #263

PR #258, opened 2023-01-04
Adding automation around ADR authoring

rogeruiz:
This patch adds support for running `npm run new-adr --slug="my-new-adr"` and
have it produce a new ADR file from the template with the suggested front-matter
from the documentation on the ADR guide page. This will help folks contributing
ADRs in the correct way without a whole lot of cognitive overhead.

### DIFF
--- a/scripts/new-adr.sh
+++ b/scripts/new-adr.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+set -e
+
+slug=$1
+
+if [ -z  "${npm_lifecycle_event}" ]
+then
+  echo "This script must be invoked by \`npm run new-adr\` and not run directly."
+  exit 1
+fi
+
+if [ -z "${slug}" ]
+then
+  echo "--slug'my-example-adr-name' needs to be passed in when running \`npm run new-adr\`"
+  exit 1
+fi
+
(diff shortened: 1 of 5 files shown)


YanZ777 (approved):
Tried it out and works for me!

scripts/new-adr.sh
```diff
@@ -0,0 +1,32 @@
...
+
+num_prefix=$(printf '%04d' "$(find docs/adrs -iname '*.md' | wc -l)")
+
+new_adr_path="./docs/adrs/${num_prefix}-${slug}.md"
+
+cat << ADRHD > "${new_adr_path}"
+---
+title: '${num_prefix} A title for this ADR'
+description: |
+  A description for ADR ${num_prefix}
```
felipe-lee, line 27: **suggestion (non-blocking)**: This doesn't seem to be part of the regular template, but I wonder if we should add it, for anyone that doesn't use this script. I would say that could be a separate PR except that would affect the `tail` command below, so it might be good to do it all in one PR.

felipe-lee (approved):
LGTM! Just left a comment, but I'm good with this whether or not that is taken into account.

scripts/new-adr.sh
rogeruiz: Thanks for the suggestion @felipe-lee I'll add it to this PR.

PR #278, opened 2023-01-26
[ MB-15044 ] Add GitHub code blocks to the documentation

rogeruiz:
- [x] Adding the GitHub Codeblock theme plugin;

Feeling inspired by @yanz777's great work on MB-15044, I found that Docusaurus
has a plugin that brings in the code from a public GitHub repository into
Docusaurus as fenced codeblock. This seems particularly helpful for the intended
reader of this kind of documentation.

Here's an example of how #277 would need to be changed to use this new
codeblock. 

```diff
diff --git a/docs/tools/cicd/github_actions/auto-approve.md b/docs/tools/cicd/github_actions/auto-approve.md
index 6da7e7f6d..d951e35f0 100644
--- a/docs/tools/cicd/github_actions/auto-approve.md
+++ b/docs/tools/cicd/github_actions/auto-approve.md
@@ -2,7 +2,15 @@
 
 ## Configuration
 
+<!--
+The following link is commented to out, but can be used if the codeblock below
+fails to render for whatever reason.
 [`auto-approve.yml`](https://github.com/transcom/mymove/blob/main/.github/workflows/auto-approve.yml)
+-->
+
+```yml reference
+https://github.com/transcom/mymove/blob/main/.github/workflows/auto-approve.yml
+```
 
 ## Purpose
 
```

And here's an image preview of the code changes from above.

Captura de pantalla 2023-01-26 a la(s) 8 26 20 a m

This branch does not need to be merged in before all the other work. We can
update the other branches after this makes it into the default branch as a nice
to have feature addition to the work in MB-15044.

[➡️  Link to Jira Ticket 🔒](https://dp3.atlassian.net/browse/MB-15044)

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -182,6 +182,7 @@ module.exports = {
     ],
   ],
   themes: [
+    '@saucelabs/theme-github-codeblock',
     [
       require.resolve("@easyops-cn/docusaurus-search-local"),
       /** @type {import("@easyops-cn/docusaurus-search-local").PluginOptions} */
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "@docusaurus/plugin-client-redirects": "2.2.0",
     "@docusaurus/preset-classic": "2.2.0",
     "@easyops-cn/docusaurus-search-local": "^0.33.6",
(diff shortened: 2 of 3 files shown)


YanZ777 (approved):
This is going to be so useful! Thanks for adding it in!

PR #294, opened 2023-02-08
[MB-15151 ]Update React Query docs

afoon:
[MB-15151](https://dp3.atlassian.net/browse/MB-15151)
This PR updates the React Query documentation. I've updated the name to Tanstack Query since that is what its called now but did leave a callout that using React with Tanstack query is still called React Query.

Updates: 
- [x] Change React Query v2 references to TanStack Query v4 references.
- [x] Code examples and best practices
- [x] Section on testing requirements

[MB-15151]: https://dp3.atlassian.net/browse/MB-15151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


afoon:
@rogeruiz tagged you for review since you were the last person to update the react query docs. I opted to add code example and more context to our documentation. This is because the official documentation is not reliable. For example The v2 links that we had before showed v3 documentation instead.

docs/frontend/guides/how-we-use-react-query.md
```diff
@@ -54,31 +88,135 @@ have callback functions that are specific to user actions. This allows us to do
 specific Component or Page actions on Success or Error related to the data being
 shown on that Component or Page.
 
+### Writing Mutations
+
+There are two mutation functions, `mutate` or `mutateAsync`. The `mutate` function does not return anything and utilizes React Query's built in error handling. The `mutateAsync` function returns a promise but requires manual error handling. Preference is given to using the `mutate` function because errors are handled.
```
YanZ777, line 93: **question:** Is that a preference that TanStack recommends or that we are suggesting?

YanZ777 (approved):
It would be nice to use code blocks, but I know that you've also added comments in the code itself for this document. 
https://docusaurus.io/docs/markdown-features/code-blocks

Thanks for updating the documentation!

docs/frontend/guides/how-we-use-react-query.md
afoon: These are TanStack's "recommendations" but not following them causes error and inconsistent behavior 🙃

YanZ777: Could you mention that in the documentation? I don't want folks to get caught off guard when they don't use the preference method.

YanZ777:
I am so sorry. I realized that I linked to the wrong thing, it's actually even more straightforward to do code blocks.
https://github.com/saucelabs/docusaurus-theme-github-codeblock

This will have the benefit of ensuring code in our documentation doesn't fall out of date, especially if it is referencing code in the mymove repository directly.

donaldthai (approved):

PR #312, opened 2023-03-09
Clarify how we use Pop's TableName method

reggieriser:
In [this milmove PR](https://github.com/transcom/mymove/pull/10224), we started including a `TableName` method on every Pop model to avoid potential issues with the way Pop tries to guess the proper table name.  The doc change in this PR attempts to clarify that we do this and why we do it.


ahobson (approved):
📝

PR #313, opened 2023-03-09
Adjust instructions for upgrading golang

reggieriser:
This PR adjust the instructions for upgrading golang based on my latest experience in [upgrading to go 1.20.x](https://github.com/transcom/mymove/pull/10185).

### DIFF
--- a/docs/backend/guides/how-to/upgrade-go-version.md
+++ b/docs/backend/guides/how-to/upgrade-go-version.md
@@ -52,11 +52,11 @@ go version go1.16.4 darwin/amd64
 
 - After your Docker image PR lands, grab the git hash from [Docker](https://hub.docker.com/r/milmove/circleci-docker) that corresponds with your merged code.
 - Update files with the updated Docker image tag hash and/or Go version:
-  - [Example PR](https://github.com/transcom/mymove/pull/9423)
+  - [Example PR](https://github.com/transcom/mymove/pull/10185)
   - Update the Go version number in:
-    - `README.md` (there's a path in the nix section that references a go version number)
     - `.go-version`
     - `.tool-versions`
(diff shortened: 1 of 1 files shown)


ahobson (approved):

docs/backend/guides/how-to/upgrade-go-version.md
```diff
@@ -67,15 +67,13 @@ go version go1.16.4 darwin/amd64
...
 - If the major/minor version changed (the first or second number, e.g. 1.x.y to 2.x.y or 1.15.x to 1.16.x):
-  - [Example PR](https://github.com/transcom/mymove/pull/4990)
+  - [Example PR](https://github.com/transcom/mymove/pull/10185)
   - Update the following files with the new go version:
-    - `go-version` in `.github/workflows/go-auto-approve.yml`
     - `go.mod`
+  - You may have to update the `golangci-lint` pre-commit hook version found in `.pre-commit-config.yaml` to one that supports the new Go version
 - Rerun the Go formatter on the codebase with `pre-commit run --all-files golangci-lint`
 - Regenerate mocks (in case the signatures have changed that we're mocking): `make mocks_generate`
 - Run `make e2e_test_docker` to test that the `Dockerfile.*local` files work  with the new image.
```
ahobson, line 79: I suppose for maximum safety, we should always deploy to experimental for go upgrades, but is that too much? I guess the only time we ever had a problem was when the base image changed, not the go version, right?

reggieriser: I might be forgetting something, but I don't think I've ever seen a problem directly attributable to a go upgrade that would have been caught by an experimental deploy.  The wildcard is that you might pick up some other (perhaps untested) merges to the main circleci-docker branch like we did with the base image.  I think I'll add in a warning about that here.

PR #315, opened 2023-03-16
Update React Query documentation

afoon:
[Jira](https://dp3.atlassian.net/browse/MB-15491)
- [ ] add more explanations
- [ ] Add more examples
- [ ] Change name back to React Query for discoverability


docs/frontend/guides/how-we-use-react-query.md
```diff
@@ -143,16 +143,60 @@ const onSubmit = async () => {
...
+  mutationFn: functionToBeCalled ,
+    // First callback
+   onSuccess: () => { 
+       console.log('Where logic should be handled') 
+       queryClient.invalidateQueries({queryKey: [myKey]})
+       },
+   onError: () => {
+       console.log('Error handling');
+       }
+   });
```
rogeruiz, line 163: Can we make these references too? I'm concerned with the code copied here being out of date going forward.

afoon: The references were moved to the section below with added context.

rogeruiz:
I'm gonna take a further look at this today and give my review. I like the intentions of these changes. This is a great start.

docs/frontend/guides/how-we-use-react-query.md
```diff
@@ -143,16 +143,60 @@ const onSubmit = async () => {
 :::
 
  Callbacks maybe not fire as expected. To avoid that issue, logic should be handled in the `useMutation` callback which is called first. 
- ```jsx reference
-https://github.com/transcom/mymove/blob/main/src/pages/Office/EditShipmentDetails/EditShipmentDetails.jsx#L22-L30
+
+```js 
+ import {useQueryClient} from '@tanstack/react-query'
```
rogeruiz, line 148:
```suggestion
```js title="use-mutation-example.jsx"
import {useQueryClient} from '@tanstack/react-query'
```

docs/frontend/guides/how-we-use-react-query.md
```diff
@@ -143,16 +143,60 @@ const onSubmit = async () => {
...
+       console.log('Error handling');
+       }
+   });
 ```
+
 UI changes should happen in the `mutate` callback which is called second after the `useMutation` callback. This is handled second so the mutation can complete. 
 
-```jsx reference
-https://github.com/transcom/mymove/blob/main/src/components/Office/ShipmentForm/ShipmentForm.jsx#L330-L339
+```js
```
rogeruiz, line 168:
```suggestion
```js title="second-callback-example.jsx"
```

rogeruiz (changes requested):
❌ This PR needs changes ...

I've added some comments on what I think we need to change. Some minor things, but otherwise this looks good for now.

rogeruiz:
Also this is unrelated to the review, but I've noticed a lot of space characters at the end of the line. The doesn't have an `.editorconfig` file, but I'll be adding one based on the `transcom/mymove` one in a new PR.

rogeruiz (approved):
lgtm 🌈

PR #317, opened 2023-03-20
[ office-hours ] Adding an update EditorConfig file

rogeruiz:
This was never added to the repository, so I copied it from
`transcom/mymove` and modified it to be more relevant for this codebase.

Discovered in #315.

The added benefit here is that anyone with EditorConfig enabled in their
IDE/PDE, will be able to trim space-characters from the end of any files they
interact with.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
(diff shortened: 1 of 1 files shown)


michaelvaldes-truss (approved):
LGTM! nice incremental improvement

PR #320, opened 2023-03-21
Added TableName to other Pop model code examples

reggieriser:
In #312, I updated some documentation about using the `TableName` method in all models and added it to one of the examples on the page.  I inadvertently left `TableName` off of the other model examples on that page, so this PR adds it to them.  I also fixed some tabs vs. spaces whitespace inconsistencies that was throwing off indentation in some markdown renderers.


mandaem (approved):
LGTM md preview

PR #325, opened 2023-03-29
MB-15567 Discovery: How does our integration with DTOD work

YanZ777:
Add new documentation about DTOD's architecture in the code and how to test it.

Information has mainly been pulled from these sources
This is an oncall incident that has useful information inside of the thread.
https://ustcdp3.slack.com/archives/CP4U2NKRT/p1677180784605049

This is a coalescence of information pulled into one page.
https://dp3.atlassian.net/wiki/spaces/~257884539/pages/2084569214/DTOD+Digging

### DIFF
--- a/docs/backend/guides/dtod/testing-dtod.md
+++ b/docs/backend/guides/dtod/testing-dtod.md
@@ -1,0 +1,47 @@
+# Testing DTOD
+
+The instructions that follow will say to change the environment variable inside of `.envrc`, however, there are other methods one can take to update to environment variables.
+
+- Environment variables can be changed inside of `.envrc.local`
+- Runing `export DTOD_USE_MOCK=false` after `direnv allow`
+
+:::note
+If you run the server in Goland, be sure to refresh the Goland environment.
+:::
+
+## Triggering the DTOD Call
+
+The following steps just documents one way in which DTOD is used in the application. There are others.
+
+1. Login as a services counselor to the office application.
+2. Create a PPM shipment.
+3. The save and continue button should trigger a call to the DTOD API.
+
+## With Mocks
+
+Ensure that `DTOD_USE_MOCK` is set to `false` in `.envrc`. Setting this to false will use mock caluclations.
+
+## Using Real DTOD
+
+Set `DTOD_USE_MOCK` in `.envrc` to `true`.
+
+In the server logs, you should see this line
+
+```
+2023-02-23T23:30:59.174Z INFO route/planner.go:167 Using real DTOD for DTOD route planner {"git_branch": "main", "git_commit": "d19fedae045252bc8f0116b1a7b216b4a1dc8927"}
+```
+
+### Changing the Password Locally
+
+The password can be overriden by running `export DTOD_API_PASSWORD=newpassword` after `direnv allow` runs.
+
+## Getting Additional Logging
+
+Inside of `pkg/route/planner.go`, change `gosoap.SoapClient` to use `gosoap.SoapClientWithConfig`
+
+### Challenges with Getting Additional Logging
+
+The soap library being used only returns an error if the status code is < 200 or >= 400.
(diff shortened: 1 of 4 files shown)


docs/backend/guides/dtod/testing-dtod.md
```diff
@@ -0,0 +1,51 @@
+# Testing DTOD
+
+The instructions that follow will say to change the environment variable inside of `.envrc`, however, there are other methods one can take to update to environment variables.
+
+- Environment variables can be changed inside of `.envrc.local`
+- Runing `export DTOD_USE_MOCK=false` after `direnv allow`
+
+:::note
+If you run the server in Goland, be sure to refresh the Goland environment.
```
felipe-lee, line 9: **note**: A plugin I've been testing out is https://plugins.jetbrains.com/plugin/19275-better-direnv and so far it seems to work for the last few times I changed the `.envrc*` files.

docs/backend/guides/dtod/testing-dtod.md
```diff
@@ -0,0 +1,51 @@
+# Testing DTOD
+
+The instructions that follow will say to change the environment variable inside of `.envrc`, however, there are other methods one can take to update to environment variables.
+
+- Environment variables can be changed inside of `.envrc.local`
+- Runing `export DTOD_USE_MOCK=false` after `direnv allow`
+
```
felipe-lee, line 7: **suggestion (non-blocking)**: A third option is to run the server like this:

```shell
export DTOD_USE_MOCK=false make server_run
```

docs/backend/guides/dtod/testing-dtod.md
```diff
@@ -0,0 +1,51 @@
...
+### Changing the Password Locally
+
+The password can be overriden by running `export DTOD_API_PASSWORD='newpassword'` after `direnv allow` runs.
+
+:::caution
+If the password contains special characters, ensure that they are escaped appropriately.
+:::
+
+## Getting Additional Logging
```
felipe-lee, line 42: **suggestion (blocking)**: I think this section should be noted as something to only do locally, not in deployed environments.

docs/backend/guides/dtod/testing-dtod.md
```diff
@@ -0,0 +1,51 @@
...
+
+Inside of `pkg/route/planner.go`, change `gosoap.SoapClient` to use `gosoap.SoapClientWithConfig`
```
felipe-lee, line 44: **suggestion (non-blocking)**: I think including this snippet could be nice so people can copy/paste:

```golang
		soapClient, err := gosoap.SoapClientWithConfig(dtodWSDL, httpClient, &gosoap.Config{Dump: true})
```

docs/backend/guides/dtod/testing-dtod.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Challenges with Getting Additional Logging
+
+The soap library being used only returns an error if the status code is < 200 or >= 400.
+https://github.com/tiaguinho/gosoap/blob/f4a99995a898b6a2de86e74d0942ffc4cfa89c0d/soap.go#L251-L261.
+
+The `Dump: true` config option should not be used as that dumps out the username and password into the logs as well.
```
felipe-lee, line 51: **suggestion**: Related to my comment above about this section probably only being for local, I don't think we need to tell people not to use the `Dump` option. In fact, it's really the only reason to change the config at all. The warning is good to keep though, to explain why this should only be done locally.

felipe-lee (commented):
Thank you for putting these docs together! This mostly lgtm, but I left a few suggestions for changes.

docs/backend/guides/dtod/testing-dtod.md
felipe-lee: Oh wait, sorry, friday afternoon brain made a mistake, the export shouldn't be here, so just

```shell
DTOD_USE_MOCK=false make server_run
```

sorry about that 😞

felipe-lee (approved):
LGTM except for one mistake in my suggestions (sorry 😞). Once that's fixed though I think this is good to go 🚢

PR #327, opened 2023-03-30
MB-15567 Discovery: How does our integration with DTOD work (The Environment Variables part)

YanZ777:
Add documentation around how environment variables get from AWS into Go.
I went with putting it under the CI/CD because it felt like that was the most relevant spot. However, if folks think it should be somewhere else, I'm open to moving it too.


docs/tools/cicd/environment_variables/README.md
```diff
@@ -0,0 +1,16 @@
+# Environment Variables
+
+To manage our environment variables, we use [direnv](https://direnv.net/) which evaluayes the `.envrc` file when changing into the repository direction.
```
felipe-lee, line 3: **suggestion**: small typos

```suggestion
To manage our environment variables, we use [direnv](https://direnv.net/) which evaluates the `.envrc` file when changing into the repository directory.
```

felipe-lee (commented):
**question**: Is this about how we use environment variables locally or in CI? I know you said it's about AWS env variables in `go`, but I'm asking if we're going for local or CI docs because I think the answer is different depending on the environment.

Like for example, this seems to be where circleci sets up access to the aws staging secrets:
https://github.com/transcom/mymove/blob/d99f31acbb6d42a4b2eb341aeab30ebc9366fc40/.circleci/config.yml#L216-L225

and then that command gets used as needed in jobs as needed, like this one:
https://github.com/transcom/mymove/blob/d99f31acbb6d42a4b2eb341aeab30ebc9366fc40/.circleci/config.yml#L1415-L1425

I could definitely be wrong though, I don't know much about how we manage secrets in circleci, I'm just guessing based on what it looks like those parts of the config are doing.

So, I think what you're documenting here is more about local dev, which means it could likely go in the backend section. There are some env vars we use on the FE, like `A11Y_AUDIT` and some playwright vars, but those aren't currently set anywhere in the `.envrc` files or loaded in chamber, so idk where those should be documented. 

Maybe since env var setup could be across the app, it could be somewhere more shared like the project layout is: https://transcom.github.io/mymove-docs/docs/about/project-layout
or maybe we need a new shared section? Idk, categorization is hard 😅

YanZ777:
> **question**: Is this about how we use environment variables locally or in CI? I know you said it's about AWS env variables in `go`, but I'm asking if we're going for local or CI docs because I think the answer is different depending on the environment.
> 
> Like for example, this seems to be where circleci sets up access to the aws staging secrets: https://github.com/transcom/mymove/blob/d99f31acbb6d42a4b2eb341aeab30ebc9366fc40/.circleci/config.yml#L216-L225
> 
> and then that command gets used as needed in jobs as needed, like this one: https://github.com/transcom/mymove/blob/d99f31acbb6d42a4b2eb341aeab30ebc9366fc40/.circleci/config.yml#L1415-L1425
> 
> I could definitely be wrong though, I don't know much about how we manage secrets in circleci, I'm just guessing based on what it looks like those parts of the config are doing.
> 
> So, I think what you're documenting here is more about local dev, which means it could likely go in the backend section. There are some env vars we use on the FE, like `A11Y_AUDIT` and some playwright vars, but those aren't currently set anywhere in the `.envrc` files or loaded in chamber, so idk where those should be documented.
> 
> Maybe since env var setup could be across the app, it could be somewhere more shared like the project layout is: https://transcom.github.io/mymove-docs/docs/about/project-layout or maybe we need a new shared section? Idk, categorization is hard 😅

Categorization is hard! 
I think I'll make a new section in the backend and put it in there since that feels most appropriate.

felipe-lee (approved):
LGTM, and thanks for moving the other existing page on vars to this new section.

felipe-lee:
Oh looks like that page that moved needs to have references to it updated. I'm seeing these:

https://github.com/transcom/mymove-docs/blob/6e79bf5b76325bdd8e42f8d0e7a65b3dc369b04d/docs/tools/docusaurus/docusaurus.md?plain=1#L208-L217
https://github.com/transcom/mymove-docs/blob/6e79bf5b76325bdd8e42f8d0e7a65b3dc369b04d/utils/redirect-backend.js#L152-L157

YanZ777:
> Oh looks like that page that moved needs to have references to it updated. I'm seeing these:
> 
> https://github.com/transcom/mymove-docs/blob/6e79bf5b76325bdd8e42f8d0e7a65b3dc369b04d/docs/tools/docusaurus/docusaurus.md?plain=1#L208-L217
> 
> 
> https://github.com/transcom/mymove-docs/blob/6e79bf5b76325bdd8e42f8d0e7a65b3dc369b04d/utils/redirect-backend.js#L152-L157

Thanks for pointing those out! I missed those when I did the initial commit.

PR #333, opened 2023-04-19
MB-15466-Update readme instructions

NamibiaTorres:
The purpose of this PR is to add sections to the README about testing and updating the documentation so that it informs engineers who may need to make changes to the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,9 @@ with
           },
 ```
 
+Remember to run `yarn start` after making this change locally.
+
+### PR Reviewers
 PR reviewers can also test changes made to the documentation by adding the path to your github repo in Redocly:
 `https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/transcom/mymove/BRANCH PATH/swagger/prime.yaml#tag/ENDPOINT PATH`
 
--- a/README.md
+++ b/README.md
@@ -99,6 +99,33 @@ Note: if you're using the Fish shell, you'll need to [complete an extra step](ht
 The site should load automatically in your browser at
 [http://localhost:4000/mymove-docs/](http://localhost:4000/mymove-docs/).
 
+## Testing Locally
+Run the local server with the following commands: `yarn install`, `yarn start`.
+
+When updating the Prime API documentation in the `mymove` repo with changes made to `prime.yaml` you can see those changes by updating `docusaurus.config.js`.
+Simple replace the following section:
+```
+          {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -107,7 +137,11 @@ Be aware that GitHub pages has a _soft_ limit of 10 deploys per hour, and it is
 
 ## API Documentation
 
-[Please read more about how Redocusaurus is being used for API documentation.](https://transcom.github.io/mymove-docs/docs/dev/tools/redocusaurus)
+[Please read more about how Redocusaurus is being used for API documentation.](https://transcom.github.io/mymove-docs/docs/dev/tools/redocusaurus).
+
+### Updating Example Data in Documentation
+The example data in the documentation is based on swagger definitions and the behavior defined in the Service and Model. Redocly will generate examples and fill in information that is not explicitly added to swagger by default.
```
ronaktruss, line 143: nit: When I read the second sentence the subject is a bit unclear what redocly will generate. I think we are trying to say that without explicit example data redocly we generate and fill in some of that information. Does something like the following feel more clear?
`When example data is not provided redocly will generate examples based on the information it has available to it like the type or enum`

README.md
```diff
@@ -107,7 +137,11 @@ Be aware that GitHub pages has a _soft_ limit of 10 deploys per hour, and it is
...
+In some cases, the response example will show all polymorphic types even if some types are not updatable. This is due to how Redocly pulls in the data to generate the examples, and it is best to add details on what is not updatable in the description of the endpoint to avoid confusion when looking at the examples.
```
ronaktruss, line 144: It's not clear to me what this is describing. Maybe adding an example would help clarify things?

ronaktruss (commented):
Thanks for adding this! This has been information that has been passed along the grapevine for a while so it's nice to finally codify it.

Preface: For nits feel free to address the ones that you actually agree with. If you're on the fence you don't need to make the changes. 

General nit: I think all of the headers should have an empty line after and code-block should have an empty line before and after. These seem to be fairly common suggestions across different markdown style guides. (If we really want to enforce these we should leverage automation and include it as part of the build process. Fwiw, your editor might have a pluggin that enables quick fixes for this. There are also cli tools for this.  Eg. https://github.com/markdownlint/markdownlint and https://github.com/DavidAnson/markdownlint.)

kctruss (approved):
💖 This additional information is really useful, thank you for adding it!

PR #334, opened 2023-04-20
Document test factory variant best practices

ahobson:
Add some clarification around factory variants


mmayo-truss (approved):
Thanks for this clarification @ahobson!

PR #338, opened 2023-04-26
MB-15819 Documentation on Modals

YanZ777:
Add documentation for how to use modals.

### DIFF
--- a/docs/frontend/guides/using-modals.md
+++ b/docs/frontend/guides/using-modals.md
@@ -101,3 +101,8 @@ Imports for MyModal in Storybook should work as follows.
 ```js
 import { MyModal } from "./MyModal";
 ```
+
+## Resources
+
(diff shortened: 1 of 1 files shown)


rogeruiz (approved):
lgtm 🌈 

This looks good. You mentioned this in the checkin meeting as well. Should remove the migrated Modal from the code base?

> At no point in time should folks be importing from MigratedModal.

YanZ777:
> This looks good. You mentioned this in the checkin meeting as well. Should remove the migrated Modal from the code base?

The `MigratedModal` should still be in the codebase as it is being used by `Modal`. It might be good for a future ticket to see if we can merge `Modal` and `MigratedModal` together though.
The idea is that most people should be importing from `Modal` instead of `MigratedModal`.

PR #340, opened 2023-04-28
Add new integrations tab

YanZ777:
Add a new tab for integrations so that they'll be easier to find instead of being buried under the backend section.

image


kctruss (approved):
This is really great, I love the reorganization and how much easier this will be to find.

PR #351, opened 2023-05-31
[ MB-15994 ] Prime API Versioning ADR

mmayo-truss:
[JIRA Ticket](https://dp3.atlassian.net/browse/MB-15994)

This ADR is addressing the technical approach for Versioning our Prime API. It is partnered with [this draft policy Doc](https://dp3.atlassian.net/wiki/spaces/MT/pages/2147188743/Draft+Prime+API+Versioning+Policy).

I have also created a test PR of possible approaches for versioning our api. I added one version 2 of the List Prime Moves endpoint. [It can be found here](https://github.com/transcom/mymove/pull/10816).

A lot of what we decide will really be more of a social compact than anything outright technical. Which approach would be best suited for how we will work together on the multiple versions of the API as a team?

### DIFF
--- a/docs/adrs/0078-api-versioning.md
+++ b/docs/adrs/0078-api-versioning.md
@@ -99,7 +99,7 @@ You can see an example of utilizing the generated code in the handlers and creat
 
 #### Services Organization
 
-This is where there are a few options to organize our services. And I am currently unsure about the best way to go.
+This is where there are a few options to organize our services.
 
 Here are some questions to consider, when adding a new endpoint to our version 2 API:
 
@@ -123,6 +123,9 @@ Here are some questions to consider, when adding a new endpoint to our version 2
 There are definitely other alternatives out there. Please feel free to suggest another.
 
 **Chosen Alternative:** Option 3. It is dependent on reliably setting the api version flag and getting it to the services. However, once this is accomplished, this option best separates the different versions.
+
+#### A Note on versioning our db/models
+While it would be possible to create versions of our db/models, we will want to avoid getting into that
 
 ### Documentation
 Communicating version changes to the prime will be crucial. We will want to deploy our prime V2 api docs
--- a/docs/adrs/0078-api-versioning.md
+++ b/docs/adrs/0078-api-versioning.md
@@ -125,7 +125,7 @@ There are definitely other alternatives out there. Please feel free to suggest a
 **Chosen Alternative:** Option 3. It is dependent on reliably setting the api version flag and getting it to the services. However, once this is accomplished, this option best separates the different versions.
 
 #### A Note on versioning our db/models
-While it would be possible to create versions of our db/models, we will want to avoid getting into that
+While it would be technically possible to create versions of our db/models, we will want to avoid that if at all possible. The complications that this would introduce into our application would be great.
 
 ### Documentation
 Communicating version changes to the prime will be crucial. We will want to deploy our prime V2 api docs
--- a/docs/adrs/0078-api-versioning.md
+++ b/docs/adrs/0078-api-versioning.md
@@ -1,5 +1,5 @@
 ---
-title: 0079 API Versioning
+title: 0078 API Versioning
 description: How to implement API Versioning
 ---
 
--- a/docs/adrs/0079-api-versioning.md
+++ b/docs/adrs/0079-api-versioning.md
@@ -32,6 +32,16 @@ Patch versions include backwards-compatible bug fixes.
 
 For example, a version could be noted as "2.1.3," where '2' is the major version, '1' is the minor version, and '3' is the patch version.
 
+#### Options on SemVer
+* Option 1: Allow for the prime to specify which minor/patch version to use
+  * Pros: Would be really granular for the consumer of our API.
+  * Cons: Would be a heavy lift to retain former minor and patch versions of our endpoints for the deprecation period.
+* Option 2: Use the most recent minor/patch versions for the respective major version, and only allow the prime to specify which major version to use.
+  * Pros: Easier to implement. For our purposes we do not need to get too complicated.
+  * Cons: Not as granular for the prime in choosing what is returned.
+
+**Chosen Alternative**: Option 2. This meets our requirements and is easier to implement. We will use the minor/patch versions to note changes in our API.
+
 ### Options for Versioning Strategy
 There are a few options for implementing api versioning in our prime API.
 
@@ -95,7 +105,6 @@ Here are some questions to consider, when adding a new endpoint to our version 2
 
 1. What do we do if we create a new endpoint that requires changes to the service that is called? How do we organize the functions/files? Do we create a new file in the services folder? Should we store these services files in a separate v2 subdir? What is the naming structure going to look like?
 2. What if the new endpoint does not require changes to the service? Should we make a new service anyway to prevent issues further down the line?
-
 
 
 * Option 1: Rename the old version to something that indicates it is old and add the new version to the same fetcher.
--- a/docs/adrs/0079-api-versioning.md
+++ b/docs/adrs/0079-api-versioning.md
@@ -11,10 +11,13 @@ description: How to implement API Versioning
 In order to support the Prime API Integration and prevent introducing potential breaking changes to the API, MilMove needs to version future updates to the Prime API. This ADR will attempt to address the "How?" questions associated with this endeavor. While some of the challenges involved are technical in nature, much of the challenge is coming up with patterns that are clear, understandable, and acceptable to the members of the engineering team.
 
 There are a few areas to consider when deciding how to implement API versioning:
-* The overall versioning strategy (URL patterns, query params, headers, etc.)
-* What endpoints will be incorporated into a new API version?
+* [The overall versioning strategy (URL patterns, query params, headers, etc.)](#options-for-versioning-strategy)
+* [What endpoints will be incorporated into a new API version?](#codebase-management)
 * An overall file organization strategy - This could even be broken up into categories to consider on their own, because we don't necessarily want to take the same organization approach for all areas of the application
-* Documentation considerations - This is a necessity since it will be key ensuring the consumers of our API are informed on what is changed between versions
+  * [Swagger files](#swagger-organization)
+  * [Handlers](#handler-organization)
+  * [Services and fetchers](#servicesfetcher-organization)
+* [Documentation considerations](#documentation) - This is a necessity since it will be key ensuring the consumers of our API are informed on what is changed between versions
 
 ## API Versioning Strategy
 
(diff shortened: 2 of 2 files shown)


ahobson (approved):
For me, this seems like a good plan. I'd love to hear from other folks if they've had other experiences

docs/adrs/0079-api-versioning.md
```diff
@@ -0,0 +1,124 @@
...
+
+Patch versions include backwards-compatible bug fixes.
+
+For example, a version could be noted as "2.1.3," where '2' is the major version, '1' is the minor version, and '3' is the patch version.
+
+### Options for Versioning Strategy
+There are a few options for implementing api versioning in our prime API.
+
+* Option 1: Include version in the URL patterns.
+  * Pros: This is the simplest option. We currently have `/v1/` in the paths of all of our APIs, and the implementation of our routing initialization can be easily modified to include multiple paths. If we went with some other option we would want to remove the `v1` from our outward facing APIs.
```
duncan-truss, line 39: I want to say this is easier to cache in some ways too, maybe more from a CDN/proxy perspective, than some of the other methods. There is a note here that gets into some of that https://learn.microsoft.com/en-us/azure/architecture/best-practices/api-design#media-type-versioning

docs/adrs/0079-api-versioning.md
```diff
@@ -0,0 +1,124 @@
...
+* Option 1: Rename the old version to something that indicates it is old and add the new version to the same fetcher.
+  * Pros: As we remove the deprecated endpoints, we will have less code duplication.
+  * Cons: It will be easier to use the wrong service with the wrong endpoint.
+  * [Example can be seen here](https://github.com/transcom/mymove/pull/10816/commits/0913f2f4625450a2ac43b358e1890d17d77b747d)
+* Option 2: Create a new fetcher and service for the new version. This can be separated out into a new file or put in the same services file. In the example linked below it is separated out into its own file.
+  * Pros: Clear delineation between the two versions. You would not be able to mix up the new service with the old service since it does not use the same fetcher.
+  * Cons: Repetitive code. Naming questions arise, in order to delineate the new from the old. Difficulty when dealing with more complicated services, especially when we use a service in another service.
+  * [Example can be seen here](https://github.com/transcom/mymove/compare/0c4be3b...3769845)
+
+There are definitely other alternatives out there. Please feel free to suggest another.
```
felipe-lee, line 110: **thought**: So, this idea just off the top of my head and tbh, I have no idea if it would actually work. What if we approached this a bit like how version managers approach things in terms of:
* something knows what version of the api/service we want (maybe a var we keep in context)
* every service has a top level entry point, e.g. `MoveTaskOrderFetcher.ListPrimeMoveTaskOrders`, which serves more as a router based on the version, and calls the appropriate version of the service
* versions of a given service are subdirectories of the service package

So for example, with the `MoveTaskOrderFetcher`, the directory structure would be something like:

```
pkg/
    move/
        move_fetcher/
            main.go    <- shell/router service which contains `MoveTaskOrderFetcher` and calls the appropriate service version
            v1/
                main.go <- contains `MoveTaskOrderFetcherV1`
            v2/
                main.go <- contains `MoveTaskOrderFetcherV2`
```

It would maybe be a bit of re-work in terms of how the services are initialized, but tbh I'm not sure. If it doesn't, this could be a pattern that can be adopted incrementally as a given service has the need to change versions. 

In this pattern, I wonder if we could somehow set the internal/ghc API request context to always have the latest version or something 🤔

docs/adrs/0079-api-versioning.md
```diff
@@ -0,0 +1,124 @@
...
+ builds off of our current routing for our API.
+
+### Codebase management
+In order to mitigate the potential for our codebase to swell in size as we branch off versions of our
+APIs, we need to address which endpoints will get pulled into the new versions.
+
+* Option 1: Pull in each endpoint as the need to implement breaking changes arises.
+  * When the first endpoint needs to be refactored to introduce a breaking change to the prime API, then a version 2 API will be created containing that one endpoint. That endpoint would be deprecated over time in version 1. More endpoints would be brought in to version 2 as the need arises.
+  * Pros: Much easier lift. Lessens the size of our codebase.
+  * Cons: Could be more confusing for the consumers of our API since not all endpoints are on each version
```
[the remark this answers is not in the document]
mmayo-truss: I have thought some about it. I think it could definitely be beneficial and would be nice to have. I am not sure about whether it would simplify our testing or not. 

However, I have thought about some possible points of confusion. Take the example where we introduce a version 2 api with a few endpoints, and then route all other endpoints to the v1 api. The prime makes the switch to version 2, after they have made the changes to their system to handle those breaking changes. Later, when we introduce a new endpoint to version 2, it will effectively be introducing a breaking change to our v2 api without incrementing the major version. While it won't technically be doing that on our end, we will just be adding a new endpoint to the v2 api, to the prime it will appear that way. They will have switched to version 2, we will then tell them to switch back to v1 for a certain endpoint because the v2 endpoint now has breaking changes. Granted, this could definitely be mitigated by clear communication.

docs/adrs/0079-api-versioning.md
```diff
@@ -0,0 +1,124 @@
...
+* An overall file organization strategy - This could even be broken up into categories to consider on their own, because we don't necessarily want to take the same organization approach for all areas of the application
+  * [Swagger files](#swagger-organization)
+  * [Handlers](#handler-organization)
+  * [Services and fetchers](#servicesfetcher-organization)
+* [Documentation considerations](#documentation) - This is a necessity since it will be key ensuring the consumers of our API are informed on what is changed between versions
+
+## API Versioning Strategy
+
+### SemVer
+The Prime API follows the SemVer (Semantic Versioning) convention, using a three-part version number: Major.Minor.Patch. Here's what each part signifies:
```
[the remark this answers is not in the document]
mmayo-truss: I can make this more explicit in the ADR. The latest minor/patch version would always be used for the respective major version. We would not be retaining minor/patch versions of endpoints. 

Thanks for calling that out.

docs/adrs/0079-api-versioning.md
mmayo-truss: @felipe-lee This suggestion is interesting. It seems like option 2, but it keeps the services even more separate and distinct. I will see if I can make a commit that illustrates this idea on my test branch.

YanZ777: **thought:** After thinking about it more, the subdir sounds like the best option to me, so similar to what Felipe recommended above. 
Having the different version of the endpoint in the same file can cause the file to get rather large and hard to navigate within the file, having a different version of the file in the same directory (without sub directorying) makes the directory cluttered and also hard to navigate. The only thing that would be problematic here would be having to be careful about imports and making sure folks who write code are importing from the correct version.

Ryan-Koch (approved):

YanZ777 (approved):
API versioning is an incredibly tricky topic. I think what you've proposed sounds reasonable to me. 
My minor note: Don't forget to remove the sections regarding your uncertainty of what option to pick, if you were planning on removing those sections. 
Thank you for taking on this ADR and putting together a draft PR as well!

docs/adrs/0079-api-versioning.md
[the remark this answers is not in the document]
duncan-truss: Ooh that's a good counterpoint scenario that I wasn't thinking of.  I might wonder if really that v1 endpoint that didn't get bumped to v2 with the others, shouldn't go straight to v3 in terms of releases of the API.

michaelvaldes-truss (approved):
Thanks for your work in creating this ADR and organizing it into readable sections -- this is a really tough thing to suddenly introduce into a mature project.
I agree with all of the chosen solutions, which generally choose to separate api versions in different directories. 
As a non-blocking suggestion, I think it might also be beneficial to include a note about how we would not plan to ever version our models/db tables.

docs/adrs/0078-api-versioning.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### Handler Organization
+
+Since we will be generating separate `*messages` and `*operations` packages for the version 2 api, it would make sense then to create a separate directory for the version 2 handlers. This will make it easier to differentiate the handlers from different versions, and ensure we are using the correct generated code for that handler.
+
+You can see an example of utilizing the generated code in the handlers and creating the necessary routing [in this commit](https://github.com/transcom/mymove/pull/10816/commits/0c4be3bab159a49f83e407830a9fecb9d27278da).
+
+#### Services Organization
+
+This is where there are a few options to organize our services. And I am currently unsure about the best way to go.
```
felipe-lee, line 102: **suggestion (non-blocking)**: I think this is one of the statements that can get removed now that there is a chosen option.
```suggestion
This is where there are a few options to organize our services. 
```

docs/adrs/0078-api-versioning.md
```diff
@@ -0,0 +1,133 @@
...
+  * Pros: Clear delineation between the two versions. You would not be able to mix up the new service with the old service since it does not use the same interface.
+  * Cons: Repetitive code. Naming questions arise, in order to delineate the new from the old. Difficulty when dealing with more complicated services, especially when we use a service in another service.
+  * [Example can be seen here](https://github.com/transcom/mymove/compare/0c4be3b...3769845)
+* Option 3: Create subdirectories in the current services directories for each version. Pass along a version flag in the AppContext (or some other method). Use that flag to identify which version of the service to use.
+  * Pros: Clear delineation between the two versions. There would not need to be a break from how we are currently utilizing our services and interfaces. We would only need to pull in the changed services into our subdirectories, which would lead to less code duplication.
+  * Cons: We would need to ensure that our flagging system worked completely as intended in order to not accidentally introduce breaking changes.
+  * [Example can be seen here](https://github.com/transcom/mymove/compare/0c4be3b...8ad4aed)
```
felipe-lee, line 121: **praise**: thank you for putting together this sample code! Really thank you for doing all the examples, but calling it out here since it was based an a suggestion late in the game. I think as you mention in the comments on the code, we'll need to figure out exactly how we want to work with the version in the context, but the example way you did it works to prove that this option can work.

felipe-lee (approved):
LGTM, thanks for all the effort you've put into putting this together and driving it forward!

docs/adrs/0078-api-versioning.md
mmayo-truss: Thanks for that @felipe-lee. I removed that.

mmayo-truss:
> Thanks for your work in creating this ADR and organizing it into readable sections -- this is a really tough thing to suddenly introduce into a mature project. I agree with all of the chosen solutions, which generally choose to separate api versions in different directories. As a non-blocking suggestion, I think it might also be beneficial to include a note about how we would not plan to ever version our models/db tables.

Thanks @michaelvaldes-truss, I added a small section about it.

PR #363, opened 2023-06-22
[ MB-16175 ] Have deploy workflow run on a schedule

rogeruiz:
This cron is set to midnight UTC in order to pull in changes from the
`transcom/mymove` repository that modify the Swagger files. Because the
docs are embedded at build time, the Docusaurus site needs to be rebuilt
and deployed in order to pick up those changes.

The need for this work was discovered when Pamplemoose was updating Swagger API
definitions and the team had not updated the Docusaurus deployment. After
discussing that it was a manual update to this repository, the team decided to
make it an automated process where a UTC midnight deployment would pull in the
latest changes in case a Truss engineer forgot to run the manual deployment
process.


kctruss (approved):
💥 Nice marginal gain! Thanks for doing this.

rogeruiz:
Here's the scheduled deployment running. 
https://github.com/transcom/mymove-docs/actions/workflows/deploy.yml?query=event%3Aschedule

It is strange to me though that this workflow ran at 20:21 ET which means it ran at 01:21 UTC. This is different than the scheduled `cron` definition of `0 0 * * *`. Could it be that GitHub Actions "scheduled runs" may not accurately run at the scheduled time? If that's true, it should be okay for what we are trying to do here. But it's useful information to know if we ever have GitHub Actions that need to run at a precise time interval.

kctruss:
> Here's the scheduled deployment running. https://github.com/transcom/mymove-docs/actions/workflows/deploy.yml?query=event%3Aschedule
> 
> It is strange to me though that this workflow ran at 20:21 ET which means it ran at 01:21 UTC. This is different than the scheduled `cron` definition of `0 0 * * *`. Could it be that GitHub Actions "scheduled runs" may not accurately run at the scheduled time? If that's true, it should be okay for what we are trying to do here. But it's useful information to know if we ever have GitHub Actions that need to run at a precise time interval.

That's frustrating, and seems to be a [known problem with GitHub Actions](https://upptime.js.org/blog/2021/01/22/github-actions-schedule-not-working/)

PR #365, opened 2023-06-26
ADR: Use flipt for feature flags

ahobson:
This builds off the work that @pbtruss did exploring feature flag options.

This PR supercedes #355


ahobson:
Thank you for your comments and questions!

> Can we get examples of how the workflow would look like if we use an option like Flipt? For example, would it change how I merge things into the base branch.

I think the workflow for all feature flag options will be approximately the same: Add the code behind the feature flag that defaults to disabled and then use the feature flag system to enable it appropriately in the various environments. We'll probably have to explore how best to do that but we can possibly/probably have ways to do things like have all users that match a certain pattern have all flags enabled / disabled. I was guessing we'd probably have most things enabled in demo and use staging as a way to test out our production rollout strategy.

> Also, I'm curious to why no frontend solutions were explored. I've work on large software that just hid the UI using feature flags managed in a database.

How we expose this to the frontend is another discussion, I think. I was imagining we would use a pattern similar to our [Restricted](https://github.com/transcom/mymove/blob/main/src/components/Restricted/Restricted.jsx) component. There are pros and cons to that approach, but again, I believe that would be shared by all feature flag approaches.

I did add a section about rolling our own feature flags and why I believe that is not as optimal.

docs/adrs/0080-use-flipt-feature-flags.md
```diff
@@ -0,0 +1,106 @@
...
+## Pros and Cons of the Alternatives
+
+### Option #0: Do Nothing
+* '+' No change to App Eng workflow
+* '+' No additional setup for Platform Team
+* '-' We will at some point ship the wrong code
+* '-' Will break API versioning required by Prime
+
+
+### Option #1: LaunchDarkly
```
felipe-lee, line 72: **suggestion (non-blocking)**: Could you add links to the respective docs for each of these options? Or more specifically, options 1, 2, 3, and 5.

docs/adrs/0080-use-flipt-feature-flags.md
```diff
@@ -0,0 +1,106 @@
...
+* '-' Adds another thing for Platform to do before Milestone 1
+* '-' Open source version may be limited
+* '-' Somewhat complicated deployment model
+
+
+### Option #3: AWS AppConfig
+* '+' Not a lot of additional setup, App Eng already used to getting other configuration information from AWS
+* '+' Very little Platform setup, AppEng already has limited access to AWS console
+* '+' Cost is rolled into existing AWS ODC, no separate approval
+* '-' Unknown cost
```
YanZ777, line 92: **question:** How are both of these statements true?

docs/adrs/0080-use-flipt-feature-flags.md
```diff
@@ -0,0 +1,106 @@
...
+* Option #4: Roll our own
+* Option #5: Flipt: Open source, self hosted, supports file based config
+
+## Decision Outcome
+
+Chosen Alternative: Option #5: Flipt
+
+LaunchDarkly is the industry standard, but there is no room for it in the budget. Unleash has similar features to AWS AppConfig, but requires more setup and maintenance effort when self-hosted (e.g. it requires a PostgreSQL database). AWS AppConfig is really a thin shim over config files and doesn't provide much in the way of helping us manage feature groupings.
+
+Rolling our own is a possibility, but one key feature we'd want to think about is how to promote feature flags from one environment to another (e.g. from staging to production). That suggests that tying the feature flag to the environment by e.g. storing the flags in the database isn't ideal as we need to recreate the flag settings in each environment. Combining that with the desire to have feature flags be enabled on a per user basis, it makes the complexity of rolling our own outweigh the relatively simple flipt deployment.
```
YanZ777, line 57: **suggestion:** I'd want to put these two paragraphs into their own respective section or with the option since I feel like it's confusing to read 
Chosen alternative: Option #5: Flip 
<content that is not about Flip>

docs/adrs/0080-use-flipt-feature-flags.md
```diff
@@ -0,0 +1,106 @@
...
+
+Flipt provides a [filesystem backend](https://www.flipt.io/docs/experimental/filesystem-backends) which would allow a way for us to manage our feature flags using a [gitops](https://about.gitlab.com/topics/gitops/) style process. We can test out our flag configuration in separate environments (e.g. experimental, demo, staging) before rolling out to production. It also allows us to deploy the Flipt service without requiring another stateful system (e.g. no database).
+
+We will have a couple of different options for how we deploy flipt. Examining the options for how flipt is deployed may be its own ADR.
```
YanZ777, line 61: **suggestion:** If this is not its own ADR, I would recommend a prototype branch. 
If it is its own ADR, then I would remove the word "may be its own" and replace it with "will be its own".

felipe-lee: I think Drew has a spike PR/branch here: https://github.com/transcom/mymove/pull/10970

docs/adrs/0080-use-flipt-feature-flags.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Pros and Cons of the Alternatives
+
+### Option #0: Do Nothing
+* '+' No change to App Eng workflow
+* '+' No additional setup for Platform Team
+* '-' We will at some point ship the wrong code
+* '-' Will break API versioning required by Prime
```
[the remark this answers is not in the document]
ahobson: I added some more text here. Maybe it helps?

docs/adrs/0080-use-flipt-feature-flags.md
```diff
@@ -0,0 +1,106 @@
...
+
+Scenario 2: Changing Authentication
+
+After we have been live for several months, the customer requires us to move from using login.gov to authenticate users to Okta. Writing and testing the code is no problem. But when we deploy it, every user now has to set up a new authentication account. Many of those users will be in the middle of moving to their next duty station when we spring this on them. Users are constantly moving, there will not be a break in the schedule where we can do the changeover.
+
+Scenario 3: Prime Freeze
+
+For 30 days before we go live, the Prime (HSA) requires a code freeze for testing. We do not want to stop deploying for 30 days, and we really don't want to stop development for 30 days. Feature flags would allow us to build (and maybe even deploy) code during the freeze and remain productive.
+
+## Considered Alternatives
```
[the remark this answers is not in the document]
ahobson: Thank you. I tried to clarify the wording around security for each of the options

docs/adrs/0080-use-flipt-feature-flags.md
```diff
@@ -0,0 +1,106 @@
...
+
+Flipt provides a [filesystem backend](https://www.flipt.io/docs/experimental/filesystem-backends) which would allow a way for us to manage our feature flags using a [gitops](https://about.gitlab.com/topics/gitops/) style process. We can test out our flag configuration in separate environments (e.g. experimental, demo, staging) before rolling out to production. It also allows us to deploy the Flipt service without requiring another stateful system (e.g. no database).
+
+We will have a couple of different options for how we deploy flipt. Examining the options for how flipt is deployed may be its own ADR.
+
+## Pros and Cons of the Alternatives
+
+### Option #0: Do Nothing
+* '+' No change to App Eng workflow
+* '+' No additional setup for Platform Team
```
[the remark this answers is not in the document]
ahobson: I tried to update the text here to be more clear. Maybe it's better?

docs/adrs/0080-use-flipt-feature-flags.md
```diff
@@ -0,0 +1,106 @@
...
+* '-' We will at some point ship the wrong code
```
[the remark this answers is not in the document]
ahobson: I tried to explain in a bit more detail. Maybe it's better?

docs/adrs/0080-use-flipt-feature-flags.md
ahobson: Thank you. I tried to reword to clarify

docs/adrs/0080-use-flipt-feature-flags.md
ahobson: Huh, that's interesting. I was trying to use this section to give an overview of why Flipt was chosen, and so that means it touches on all of the options, not just flipt.

docs/adrs/0080-use-flipt-feature-flags.md
```diff
@@ -0,0 +1,106 @@
...
+
+# Flipt Feature Flags to enable CI/CD and roll-forward recovery
+
+**User Story:** MB-13752
+
+Why do this?
+
+In production, USTC is going to expect very fast turnaround on some changes (hopefully mostly bugs). As they also have a long list of features to be added after go-live, it is likely that an incomplete feature will be committed when a fix is needed, and completing that feature would add an unacceptable delay to shipping the fix.
+
+There are a few different approaches to solving this problem, depending on which decade you’re in. During the 2020s, the industry generally uses feature flags, so that is the type of solution considered here.
```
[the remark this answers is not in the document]
ahobson: I've tried to add some more information and comments. Let me know what you think

brandonlenz (approved):
@ahobson thanks for presenting about this earlier today.

Just gave it another read through, and I'm convinced! You clearly put _a lot_ of thought and energy into this (especially with the example branch!). It'll take everyone a bit of time getting to know this tool, I'm sure, but that's an exciting prospect!

One language nitpick below.

docs/adrs/0080-use-flipt-feature-flags.md
YanZ777: Ohhh I see, using this section in this manner makes sense. I think it was a bit confusing for me when I read the heading / subheading and then the content, but if other folks are ok with this, then lets leave the section organized as is.

YanZ777 (approved):
The provided options and provided prototype PR make sense to me. Thanks for taking the time to write this out and solicit feedback from folks.

docs/adrs/0080-use-flipt-feature-flags.md
ahobson: I've added a sentence at the beginning of this section to indicate it is a summary of the decision. Maybe that's better?

docs/adrs/0080-use-flipt-feature-flags.md
ahobson: I've updated the working to "will be". Thank you!

brandonlenz:
@ahobson I merged ADR 0081 (expecting yours to be merged first since it got approval first). Just noting in case you're worried about needing to bump your ADR number, you won't need to, you can stick to 0080 without conflict.

docs/adrs/0080-use-flipt-feature-flags.md
```diff
@@ -0,0 +1,164 @@
...
+* '-' Very low level
+
+### Option #4: Roll our own
+* '+' Provides all flexibility we want
+* '+' Easy to deploy (probably?) and meets MilMove security requirements
+* '-' No easy way to migrate settings from one environment to another
+* '-' We have to implement all the flexibility
+
+### Option #5: [Flipt](https://www.flipt.io)
+* '+' Provides flexibility to enable feature flags to groups of users in almost any configuration we could imagine
```
duncan-truss, line 160: Do we know yet how we might cross reference a user to their segment or flipt configuration from say a logging perspective?  Will we add some new fields to the context to understand what was active at a certain time?

ahobson: This is a good question. We will have to see what we need in practice. I could see logging the response for every feature flag query (which would also have the `session_id`) which would seem to provide a way to work out what happened.

We might also find on the server side we want to do something like

```
if flag == "thing_one" {
  logger.Info("Using thing_one flag")
} else {
  logger.Info("Not using thing_one flag")
}

Client side it's a bit tricker, huh?

PR #370, opened 2023-06-30
[ MB-16540 ] [ ADR ] Deprecating devseed data scenarios

rogeruiz:
This was brought up in planning for the Pamplemoose team and also in the BE/FE
weekly check-in meeting. The consensus reached in that meeting was that
`devseed` scenarios are not as useful as `testharness` scenarios. The goal of
this ADR is to document the consensus along with a plan for deprecating the
`devseed` scenarios. This ADR will also take on the work of documenting the
decision across Docusaurus and Confluence as well.

The goal of this ADR is not to improve the `testharness` scenarios. That will be
a quick-follow ADR where we can discuss the UX and UI for the
`testharness/list`. In the next ADR, we can also explore improvements for
training and documentation that this ADR will introduce.

There's a [draft PR](https://github.com/transcom/mymove/pull/11051) with the removed devseed code that can be ephemerally deployed to get a feel for what it would be like to remove the devseed data.


mlm55:
if we're considering deleting current devseed code, should look at what uses `generate-test-data`

docs/adrs/0081-deprecating-devseed-scenarios.md
```diff
@@ -0,0 +1,110 @@
...
+- `-` _Teams will remain unsure about whether to use `devseed` or `testharness`
+  data._
+- `-` _Teams may duplicate work as `testharness` data is required for Playwright
+  tests._
+
+### _Deprecate `devseed` scenarios and provide team-wide guidance on exclusively using `testharness` scenarios_
+
+- `+` _Teams will be able to focus on data creation using `testharness` data._
+- `+` _Teams will have `testharness` data for Playwright tests that have been
+  used during the development of features._
```
rogeruiz, line 87:
```suggestion
- `+` _Teams will have `testharness` data for Playwright tests that have been
  used during the development of features._
- `+` _Removing `devseed` scenarios may drastically improve the deployment of ephemeral deploys._
```

docs/adrs/0081-deprecating-devseed-scenarios.md
```diff
@@ -0,0 +1,110 @@
...
+  - `?` _Teams will be able to create more specific move data faster._
+
+### _Deprecate and delete `deveseed` scenarios_
+
+- `+` _Teams will be able to focus on data creation using `testharness` data._
+- `+` _Teams will have `testharness` data for Playwright tests that have been
+  used during the development of features._
+- `-` _More work involved in implementing the ADR._
+- _[...]_ <!-- numbers of pros and cons can vary -->
+
```
rogeruiz, line 104: This may be the better solution here. If there is nothing that relies on `devseed` existing, then we can get rid of it pretty quickly. The only thing that uses `devseed` data is the ephemeral deployment.

The biggest lifts here are the documentation updates around `devseed`.

rogeruiz: This would also get rid of certain `make` targets related to populating the database.

michaelvaldes-truss (approved):
I think this is a great final push towards migrating fully to the testharness.
One concern I have is with being able to very quickly create a scenario, locally, where we have many moves. The devseed data makes it easy to make the queues feel like they might in production. Do we have a testharness scenario that creates many moves (or at least runs all of the other testharness scenarios)?

ronaktruss:
I think we should eventually make the switch over to Option 3 (deprecating and deleting the dev seed data), though atm I think I'd prefer something closer to Option 2, but with the caveat that we don't continue to actively support dev seed data.

With M1 approaching, if we don't have parity between the existing testharness functions and the existing devseed scenarios we are potentially creating additional work that may detrimentally impact upcoming work. We can also utilize this time between "deprecating" and deleting the code to identify scenarios that we need to add to the testharness.

ruizajtruss (approved):
I'd echo Ronak's comment. Something between option 2 and 3 seems ideal in the interim. We'd want to make sure we cover all scenarios with testharness data that devseed covers already while cutting off dev seed maintenance. Once we're sure that's been done, then it'd be a good time to start deleting all the devseed stuff and go full bore like option 3 states.

mmayo-truss (approved):
I agree with @ronaktruss, option 3 should be our eventual aim, but option 2 seems like our best way to begin this process. We can start with deprecating the devseed scenarios with a push towards eventual deletion once we ensure the testharness data covers current devseed data. This could also be an incremental process, transitioning subscenarios a few at a time from devseed to testharness. 

I think it will also be important to communicate this broadly across the project, not just AppEng, since it will change how folks will use our ephemeral deployments for testing. Perhaps a technical sharing in an all-together?

rogeruiz:
Thanks for the feedback everyone. This will get merged in on Wednesday.

PR #376, opened 2023-07-11
Updating the `asdf` troubleshooting documentation

rogeruiz:
This was brought up by @ruizajtruss while troubleshooting his
environment. These changes help get unstuck and are more generic across
different Apple chip installations. There is also an updated comment
which is explicit about the order of operations when it comes to loading
`asdf.sh` and the subsequent `go env` commands.


ruizajtruss (approved):
This is great, thanks again!

PR #383, opened 2023-07-17
MB-16460: Add link to prime v2 docs

ronaktruss:
Added a link to the v2 API docs and added a second link to the v1 docs for url consistency.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -179,6 +179,7 @@ module.exports = {
             spec: 'https://raw.githubusercontent.com/transcom/mymove/main/swagger/ghc.yaml',
             route: '/api/ghc',
           },
+          // Continue to support existing url for those that have bookmarked this link. We can route this to the oldest supported definition
           {
             spec: 'https://raw.githubusercontent.com/transcom/mymove/main/swagger/prime.yaml',
             route: '/api/prime',
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -184,6 +184,14 @@ module.exports = {
             route: '/api/prime',
           },
           {
+            spec: 'https://raw.githubusercontent.com/transcom/mymove/main/swagger/prime.yaml',
+            route: '/api/prime/v1',
+          },
+          {
+            spec: 'https://raw.githubusercontent.com/transcom/mymove/main/swagger/prime_v2.yaml',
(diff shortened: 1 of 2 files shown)


docusaurus.config.js
```diff
@@ -183,6 +183,14 @@ module.exports = {
             spec: 'https://raw.githubusercontent.com/transcom/mymove/main/swagger/prime.yaml',
             route: '/api/prime',
           },
+          {
+            spec: 'https://raw.githubusercontent.com/transcom/mymove/main/swagger/prime.yaml',
+            route: '/api/prime/v1',
```
mmayo-truss, line 188: Suggestion: You added this `/api/prime/v1` route, but left the prime route that references the same yaml file. You are referencing the prime route and not your new route. Should we just have one route for the `prime.yaml` file?

sidebars.js
```diff
@@ -89,6 +89,11 @@ module.exports = {
           label: 'Prime',
```
mmayo-truss, line 89: Question: Should this say Prime V1 now?

docusaurus.config.js
ronaktruss: My thought process was that it would be nice if we had the same pattern for the URLs. Since we now have a /v2 I thought adding a /v1 would be nice. At the same time I didn't want to cause issues for folks that had the existing link bookmarked. I think we can continue to leverage this default url to point to the main version.

mmayo-truss (approved):
I added some non blocking questions/suggestions. 

However, this works as intended. I am able to see the new link to the prime v2 api documentation when running locally. It is displaying the appropriate swagger file with the newly versioned moveTaskOrder endpoint. 

Approved.

docusaurus.config.js
mmayo-truss: @ronaktruss That totally makes sense to me. Good thinking

sidebars.js
ronaktruss: Yeah that's a good ideal.

ronaktruss: I ended up making the Prime V1 link point to the /v1 url now. https://github.com/transcom/mymove-docs/pull/383/commits/9a54bcfcfcfa3fdf2d65da0ad0e7f79dc4d72b8d

PR #384, opened 2023-07-18
MB-16460: Update ADR based on implementation PR

ronaktruss:
# Summary

This adds a new ADR indicating the findings and shifts from our [original ADR](https://transcom.github.io/mymove-docs/docs/adrs/api-versioning) based on the work done in [this PR](https://github.com/transcom/mymove/pull/10952).

I've updated the existing ADR to include a link to the new ADR, fixed a typo, and updated the formatting. There have been no material changes to the ADR itself. (This imo is immutable in spirit, but if folks feel differently please let me know.) It doesn't appear as though we have a standard way to indicate the status of an ADR. That improvement is out of the scope for this PR. But happy to hear opinions if folks feel strongly. For now I have added a `Note` and included this information at the beginning of the ADR.

I've added a new ADR referencing the old ADR and include a summary of the major differences. I've tried to make this a standalone ADR by adding some additional context, but imo it still feels a little odd since we didn't really evaluate things from first principles here. Because this new ADR isn't doing a first principles analysis it does deviate a bit from the traditional structure of an ADR.

I'm open to feedback on how to best structure this information. If its easier to make change on this branch directly or you want to make a fork and make modifications feel free to do so.

I'd be interested in getting answers to the following:
* Does this ADR structure make sense?
* Does this answer the WHY we deviated from the originally suggested approach?


afoon:
**suggestion**: Can this new ADR present the previous chosen options in ADR 78 as Option 1 and the new changes as an Option 2. I found referencing the of option numbers that are ADR 78 a little confusing. It may help to treat the ADR 78 as if it was written 2 years ago and ADR 80 is introducing a few updates.

ronaktruss:
@afoon I was originally planning on doing something similar, but not all of the sections had a "new" option. Only one of the 3 sections did. Because of that, I ended up deviating from that.

Just to be explicit, would your suggestion here be to rewrite all of the sections so we explicitly call out the existing chosen option and then not necessarily have have an option 2. (When the option 2 doesn't exist, we would instead just have the comments for that section)

afoon:
> Just to be explicit, would your suggestion here be to rewrite all of the sections so we call out the existing chosen option and then not necessarily have have an option 2. (When the option 2 doesn't exist, we would instead just have the comments for that section)

I would write only about the relevant changes. I would not mention the sections that remained unchanged because they are already cannon.

docs/adrs/0080-api-versioning-based-on-test-run.md
```diff
@@ -0,0 +1,49 @@
...
+
+This is where the largest deviation is.
+
+Previously we were suggesting to go with [Option 3: Creating subdirectories and using flags](https://transcom.github.io/mymove-docs/docs/adrs/api-versioning#services-organization). This approach had the following pros and cons:
+
+> Pros: Clear delineation between the two versions. There would not need to be a break from how we are currently utilizing our services and interfaces. We would only need to pull in the changed services into our subdirectories, which would lead to less code duplication.
+>
+> Cons: Repetitive code. Naming questions arise, in order to delineate the new from the old. Difficulty when dealing with more complicated services, especially when we use a service in another service.
+
+Through this test run, we realized that the flagging system had an additional con in that it was obfuscating complexity and enabling folks to potentially misuse the flag. The default behavior also made it so that all existing tests that used a given service would have been updated to leverage the new versioned service.
```
YanZ777, line 42: **questions:** 
What do you mean by default behavior here? 
Can you clarify what you mean by? 
> all existing tests that used a given service would have been updated to leverage the new versioned service.

ronaktruss: Thanks for calling this out. I'll try to make this more clear in the ADR as well.

If you take a look at the [example](https://github.com/transcom/mymove/compare/0c4be3b...8ad4aed) in the ADR for this choice and look at the `pkg/services/move_task_order/move_task_order_fetcher.go` file you will see that the `ListPrimeMoveTaskOrders` function tries to call the v1 or v2 service based on the status of the flags, but defaults to calling the v2 service if it fails both of the conditionals. For tests, the way that the flag was setup meant that we would always fail both of the conditionals and call the v2 service. 

Did that help clarify things?

ronaktruss: I modified this portion to be: https://github.com/transcom/mymove-docs/pull/384/commits/d68616e32f937ed126d7fcbe2895c1d795133196. Let me know if you think this clarifies things.

docs/adrs/0080-api-versioning-based-on-test-run.md
```diff
@@ -0,0 +1,49 @@
...
+
+Everything in the [previous ADR](./0078-api-versioning.md) outside of the file organization strategy is accurate and reflects the process that we follow for API versioning.
+
+## File/Code Organization
+
+### Swagger Organization
+
+We are continuing to leverage [Option 2: Have an entirely separate version 2 swagger file](https://transcom.github.io/mymove-docs/docs/adrs/api-versioning#swagger-organization) from the original ADR.
+
+Going through this approach we were able to confirm the exist pros and cons.
```
mmayo-truss, line 20:
```suggestion
Going through this approach we were able to confirm the existing pros and cons.
```

docs/adrs/0080-api-versioning-based-on-test-run.md
```diff
@@ -0,0 +1,49 @@
...
+
+Because of these findings we decided to go with a new option.
+
+Option 4: Only create new services if they are needed and shift any shared logic into reusable functions.
```
mmayo-truss, line 46: Suggestion: I feel like the differentiation should be more clear here. Maybe make it more explicit that we are not going to version services with this option?

ronaktruss: How does this sound?

```suggestion
Option 4: Only create new services if they are needed and shift any shared logic into reusable functions. These new services won't be versioned and will be added like existing functions. (Eg. if we have a service for fetching orders we might create a separate `fetchOrdersWithAddresses` function that has the same core functionality but also fetches all of the child addresses.)
```

ronaktruss:
@afoon Fysa, I I'm tentatively leaning towards not making the change you suggested. But if others also feel strongly about that suggestion I'm happy to make that change.

(My perspective: I think this ADR builds on top of the existing ADR so both documents should be viewed in tandem. If that is the case, reusing options or making it so option 1 here reflects option 3 in the first document might also be confusing.)

afoon:
> @afoon Fysa, I'm tentatively leaning towards not making the change you suggested. But if others also feel strongly about that suggestion I'm happy to make that change.
> 
> (My perspective: I think this ADR builds on top of the existing ADR so both documents should be viewed in tandem. If that is the case, reusing options or making it so option 1 here reflects option 3 in the first document might also be confusing.)

While the ADRs are related, this one does change a decision made in the previous one. The precedence of existing ADRs such as [ADR 0003](https://transcom.github.io/mymove-docs/docs/adrs/go-path-and-project-layout) /[ADR 0037](https://transcom.github.io/mymove-docs/docs/adrs/go-path-and-project-layout-revisited) and [ADR 0071](https://transcom.github.io/mymove-docs/docs/adrs/move-history-events)/[ADR 0074](https://transcom.github.io/mymove-docs/docs/adrs/move-history-details-refactor) treat the prior ADR as the existing method being changed in the new ADR. 
We have the context the this ADR was written within days of last one as a formality, but new engineers to the project will not have that context.

ahobson (approved):
Thank you for writing this up and incorporating all the feedback!

mmayo-truss (approved):
Thanks for clarifying that section on the new direction we are taking wrt our services.

PR #401, opened 2023-08-16
MB-16873 - Add documentation for server integration tests using middleware.

scottmries:
This adds documentation according to [this ticket](https://dp3.atlassian.net/browse/MB-16873), and based on existing integration tests that follow some of this pattern. 

Instructions for building the request body are based on conclusions reached in A-Team's meandering sync and the BE/FE check-in: prefer the use of existing types to leverage Intellisense and compile-time errors, as opposed to using raw strings as JSON or locally-typed objects.

### DIFF
--- a/docs/backend/testing/running-integration-tests-with-routing-middleware-enabled.md
+++ b/docs/backend/testing/running-integration-tests-with-routing-middleware-enabled.md
@@ -13,7 +13,7 @@ Your test will live in `pkg/handlers/routing`, in a subdirectory corresponding t
 The name of the test file should correspond to the name of the file that contains the handler that the endpoint calls, `{handler_file_name}_test.go`. `pkg/handlers/internalapi/ppm_shipment.go` would have a corresponding test file in `pkg/handlers/routing/internalapi_test/ppm_shipment_test.go`.
 
 Generally, the test name and location should be
-```pkg/handlers/routing/{api_name}api_test/{handler_name}_test.go```
+```pkg/handlers/routing/{api_name}api_test/{handler_file_name}_test.go```
 
 ## Test structure
 
--- a/docs/backend/testing/running-integration-tests-with-routing-middleware-enabled.md
+++ b/docs/backend/testing/running-integration-tests-with-routing-middleware-enabled.md
@@ -1,0 +1,105 @@
+---
+sidebar_position: 8
+---
+
+# Running integration tests with routing middleware enabled
+
+Sometimes it will be necessary to write server integration tests with routing middleware enabled. For example, we want to test calling an API endpoint with an unauthorized user, and ensure that we receive a 403 response. Here's how to go about it.
+
+## Test name and location
+
+Your test will live in `pkg/handlers/routing`, in a subdirectory corresponding to the API you're testing, `{api_name}api_test`. Tests of the `internal` API live in `pkg/handlers/routing/internalapi_test`.
+
+The name of the file should correspond to the handler that the endpoint calls, `{handler}_test.go`. `pkg/handlers/internalapi/ppm_shipment.go` would have a corresponding test file in `pkg/handlers/routing/internalapi_test/ppm_shipment_test.go`.
+
+Generally, the test name and location should be
+```pkg/handlers/routing/{api_name}api_test/{handler_name}_test.go```
+
+## Test structure
+
+In general, here's how to set up the file created above.
+
+Include the test as part of the `{api_name}api_test` package, and any imports, as usual.
+
+For each request type, write a function whose receiver is the API's test suite. For example, tests of updating a pro-gear weight ticket in the internal API's suite would look like:
+
+```
+func (suite *InternalAPISuite) TestUpdateProgearWeightTicket() {
+  ...
+}
+```
+
+For each test, call `Run` on the suite, with a description of the test, and the test function.
+
+```
+  suite.Run("Unauthorized progear weight ticket update by another service member", func() { ... })
+```
+
+The body of the test function will likely contain some data set up -- using factories to build a pro-gear weight ticket and service member, for example -- which we can use to build the endpoint, body, and request, and get the response.
+### Endpoint
+
+Build the endpoint from the corresponding Swagger definition. For example, the `swagger-def/internal.yaml` defines the update path for pro-gear weight tickets as
+
+```
+/ppm-shipments/{ppmShipmentId}/pro-gear-weight-tickets/{proGearWeightTicketId}
+
+```
+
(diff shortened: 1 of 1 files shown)


docs/backend/testing/running-integration-tests-with-routing-middleware-enabled.md
```diff
@@ -0,0 +1,105 @@
+---
+sidebar_position: 8
+---
+
+# Running integration tests with routing middleware enabled
+
+Sometimes it will be necessary to write server integration tests with routing middleware enabled. For example, we want to test calling an API endpoint with an unauthorized user, and ensure that we receive a 403 response. Here's how to go about it.
```
felipe-lee, line 7: **suggestion (non-blocking)**:

```suggestion
Sometimes it will be necessary to write server integration tests with routing middleware enabled. For example, when we want to test calling an API endpoint with an unauthorized user, and ensure that we receive a 403 response. Here's how to go about it.
```

docs/backend/testing/running-integration-tests-with-routing-middleware-enabled.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Test name and location
+
+Your test will live in `pkg/handlers/routing`, in a subdirectory corresponding to the API you're testing, `{api_name}api_test`. Tests of the `internal` API live in `pkg/handlers/routing/internalapi_test`.
+
+The name of the file should correspond to the handler that the endpoint calls, `{handler}_test.go`. `pkg/handlers/internalapi/ppm_shipment.go` would have a corresponding test file in `pkg/handlers/routing/internalapi_test/ppm_shipment_test.go`.
```
felipe-lee, line 13: **suggestion (non-blocking)**: Mainly making a distinction between the handler name and the name of the file containing the handler, since they may be different.

```suggestion
The name of the test file should correspond to the name of the file that contains the handler that the endpoint calls, `{handler_file_name}_test.go`. `pkg/handlers/internalapi/ppm_shipment.go` would have a corresponding test file in `pkg/handlers/routing/internalapi_test/ppm_shipment_test.go`.
```

docs/backend/testing/running-integration-tests-with-routing-middleware-enabled.md
```diff
@@ -0,0 +1,105 @@
...
+
+Generally, the test name and location should be
+```pkg/handlers/routing/{api_name}api_test/{handler_name}_test.go```
```
felipe-lee, line 16: **suggestion (non-blocking)**: related to the suggestion I made above

```suggestion
Generally, the test name and location should be
```pkg/handlers/routing/{api_name}api_test/{handler_file_name}_test.go```
```

docs/backend/testing/running-integration-tests-with-routing-middleware-enabled.md
```diff
@@ -0,0 +1,105 @@
...
+
+For each test, call `Run` on the suite, with a description of the test, and the test function.
+
+```
+  suite.Run("Unauthorized progear weight ticket update by another service member", func() { ... })
+```
+
+The body of the test function will likely contain some data set up -- using factories to build a pro-gear weight ticket and service member, for example -- which we can use to build the endpoint, body, and request, and get the response.
+### Endpoint
+
```
felipe-lee, line 40: **thought**: Your example shows this, but I think it would be good to explicitly call it out. When setting up the endpoints for these types of tests, you need to prefix the endpoint path with `/internal`, `/ghc/v1`, `/prime/v1`, etc. This is a difference from the endpoint paths that we usually set up in our handler server and integration tests that live next to the handler files because those omit that part of the endpoint.

docs/backend/testing/running-integration-tests-with-routing-middleware-enabled.md
```diff
@@ -0,0 +1,105 @@
...
+```
+/ppm-shipments/{ppmShipmentId}/pro-gear-weight-tickets/{proGearWeightTicketId}
+
+```
+
+Build the endpoint path, for example, assuming the `ppmShipment` and `proGearWeightTicket` are created,
+
+```
+endpointPath := fmt.Sprintf("internal/ppm-shipments/%s/pro-gear-weight-tickets/%s", ppmShipment.ID.String(), progearWeightTicket.ID.String())
```
felipe-lee, line 51: **issue**: leading forward slash was missing

```suggestion
endpointPath := fmt.Sprintf("/internal/ppm-shipments/%s/pro-gear-weight-tickets/%s", ppmShipment.ID.String(), progearWeightTicket.ID.String())
```

docs/backend/testing/running-integration-tests-with-routing-middleware-enabled.md
```diff
@@ -0,0 +1,105 @@
...
+
+jsonBody, err := json.Marshal(body)
+
+suite.FatalNoError(err)
+
+bodyBuffer := bytes.NewBuffer(jsonBody)
+```
+### The request
+
+Now we can build the request from the endpoint and body. Since our request is hitting the customer side of the application, we'll use an authenticated MilMove request, and a `PATCH` method. We will also set the `Content-Type` header.
```
felipe-lee, line 78: **thought**: You mention an authenticated MilMove request here and show it in the example, which is great. I wonder if we should also mention the other possible request types like the unauthenticated version, or the office and prime versions:

https://github.com/transcom/mymove/blob/d491f8db417e9d3bb8711fffba47fb2aa0163b79/pkg/handlers/routing/base_routing_suite.go#L221-L307

What do you think?

felipe-lee (approved):
Thanks for getting these docs up so quickly! I have a few suggestions, but this overall lgtm! I think it would be good to ask someone that doesn't have as much experience with this to see if it makes sense to them though.

docs/backend/testing/running-integration-tests-with-routing-middleware-enabled.md
scottmries: Good suggestion, I updated it to "we may want."

melissathai (approved):

docs/backend/testing/running-integration-tests-with-routing-middleware-enabled.md
scottmries: Added, nice suggestion.

docs/backend/testing/running-integration-tests-with-routing-middleware-enabled.md
scottmries: Yeah, that makes sense. I added reference to the other versions.

kingishb-truss (approved):
Hey, this looks really clear to me! Nice work

PR #403, opened 2023-08-18
Update GEX documentation

YanZ777:
https://dp3.atlassian.net/browse/MB-12698
Document was in draft and has now been renamed, so the URL needs to be updated.


ahobson (approved):
Link works for me

PR #479, opened 2024-02-24
Update sidebars.js

r-mettler:
add playwright to toolbar


landan-parker-0 (approved):
let's see if it works

PR #494, opened 2024-05-01
updated docs for frontend feature flag creation

loganwc:
removed old documentation for creating a function to check FF in featureFlag.js because there's a function to check any feature flag now.


danieljordan-caci (approved):

deandreJones:
thanks for this update!

